PR #25, opened 2022-11-30
[SID-594] Sign up form setup

ikovic:
https://todo.irdcorp.dev/issue/SID-594

Prep work for the new React form component.


packages/react/src/components/form/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { Success } from "./success";
+
+export const Form = () => {
+  const flowState = useFlowState();
+
+  return (
+    <div className={themeClass}>
+      <div className={styles.form}>Form</div>
+      <p>Flow state: {flowState.status}</p>
+      {flowState.status === "initial" && <Initial flowState={flowState} />}
```
ikovic, line 16: @vincenzoiozzo State based conditional rendering - makes it easy to figure out which UI should be rendered in the current state.

packages/react/src/components/form/initial.tsx
```diff
@@ -0,0 +1,25 @@
+import { InitialState } from "./flow";
+
+type Props = {
+  flowState: InitialState;
```
ikovic, line 4: @vincenzoiozzo This component can only be rendered in the `InitialState` so it expects the `flowState` to be of that type. No further checks are needed, eg in that state `.logIn` must be defined.

packages/react/src/components/form/useFlowState.ts
```diff
@@ -0,0 +1,23 @@
+import { useEffect, useRef, useState } from "react";
+
+import { useSlashID } from "../../hooks/use-slash-id";
+import { Flow, createFlow, FlowState } from "./flow";
+
+export function useFlowState() {
```
ikovic, line 6: This is used to synchronize the React components state with the flow state machine. I'll add that as a JSDoc comment here.

packages/react/src/theme/theme.css.ts
```diff
@@ -0,0 +1,10 @@
+import { createTheme } from "@vanilla-extract/css";
+
+export const [themeClass, theme] = createTheme({
```
ikovic, line 3: Later this will be used to expose the theming API for our customers.

packages/react/vitest.config.ts
```diff
@@ -1,6 +1,9 @@
 import { defineConfig } from "vitest/config";
+import { vanillaExtractPlugin } from "@vanilla-extract/vite-plugin";
 
 export default defineConfig({
+  // @ts-ignore
```
ikovic, line 5: I'll get to the bottom of this, it compiles after ignoring the error but it looks like there is some mismatch between `vitest` and `vanilla-extract`

packages/react/src/components/form/flow.ts
```diff
@@ -0,0 +1,167 @@
+import { Cancel, LogIn, LoginOptions, Retry } from "../../domain/types";
+
+export interface InitialState {
+  status: "initial";
+  logIn: (options: LoginOptions) => void;
+}
+
+export interface AuthenticatingState {
```
ikovic, line 8: @vincenzoiozzo Each state is referenced in the `FlowState` union type. `status` is used as the discriminating property so we can easily check which type we're looking at based on that value.

States have context and actions - context is the related data we guarantee to exist at that time, while actions are the functions that can be called while in that state.

On the consumer side this makes it easy to call the proper methods, because after testing the state type using `status` you can safely call these actions - TypeScript will tell you what is available.

PR #46, opened 2023-02-02
[SID-678] Bundle fonts in Form component

kaspermroz:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-678)

We want to bundle font CSS files within actual <Form /> component instead of dev preview. 🐛

## Checklist

### DIFF
--- a/packages/react/src/components/form/form.test.tsx
+++ b/packages/react/src/components/form/form.test.tsx
@@ -1,5 +1,5 @@
 import { User } from "@slashid/slashid";
-import { render, screen } from "@testing-library/react";
+import { render, screen, waitFor } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { Form } from ".";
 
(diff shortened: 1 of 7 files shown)


ikovic:
@kaspermroz can you check the ticket description for details please? Some code should be moved around to get this done 😃

packages/react/src/components/form/form.test.tsx
```diff
@@ -57,9 +57,11 @@ describe("#Form", () => {
     );
     user.click(cancelButton);
 
-    await expect(
-      screen.findByTestId("sid-form-initial-state")
-    ).resolves.toBeInTheDocument();
+    await waitFor(() =>
```
ikovic, line 60: This won't solve the problem - `findByTestId` internally already uses `waitFor`. Problem is with the test setup that is related to timers. We can solve that in a separate PR in order to keep this one clean.

ikovic (approved):
Looks good, thank you!

PR #48, opened 2023-02-03
[SID-630] Input validation

kaspermroz:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-630)

Currently we're not doing any validation in the `<Form />` component, meaning that we're allowing submission of empty and invalid data. To prevent it, we want to add basic validation for different Handle types.

## Checklist


packages/react/src/components/form/validation.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    });
+  });
+
+  it("should be invalid email address", () => {
+    invalidEmails.forEach((email) => {
+      expect(isValidEmail(email)).toBe(false);
+    });
+  });
+
+  it("should be valid OTP code", () => {
```
kaspermroz, line 88: These tests might be an overkill but I figured I might as well cover these functions in case we'll want something more elaborate in the future 🤷‍♂️

ikovic:
@kaspermroz in general the approach looks good to me. We could maybe spend some time thinking about ways of reducing the duplication? I'm thinking about those `useState` references for dirty and error states. There are different levels on which we could solve this - a single hook that does the job for us, a `Form` component with some `React.Context` magic (input fields could read the error from context and render it automatically).

We don't necessarily need to implement any of this, I'd just like to evaluate the trade offs 🤓

packages/react/src/components/form/initial.tsx
```diff
@@ -113,12 +118,16 @@ const HandleForm: React.FC<HandleFormProps> = ({
...
-  const [email, setEmail] = useState("");
-  const [phone, setPhone] = useState("");
+  const { registerField, registerSubmit, values, status, resetForm } =
+    useForm();
   const [flag, setFlag] = useState<Flag>(GB_FLAG);
   const [factor, setFactor] = useState<Factor>(filteredFactors[0]);
   const { text } = useConfiguration();
 
+  useEffect(() => {
+    return resetForm;
```
kaspermroz, line 128: cleanup when changing tabs so we don't have stale errors

ikovic: Nice!

packages/react/src/context/form-context.tsx
```diff
@@ -0,0 +1,130 @@
...
+        setValues((v) => ({ ...v, [fieldName]: "" }));
+      }
+
+      if (!validators[fieldName] && validator) {
+        setValidators((v) => ({ ...v, [fieldName]: validator }));
+      }
+
+      return (e) => {
+        setValues((v) => ({ ...v, [fieldName]: e.target.value }));
+        if (validator && status === "invalid") {
```
ikovic, line 72: Nice!

packages/react/src/components/form/form.test.tsx
```diff
@@ -21,6 +27,42 @@ describe("#Form", () => {
     expect(screen.getByTestId("sid-form-initial-state")).toBeInTheDocument();
   });
 
+  test("should render error message - empty email input", async () => {
```
ikovic, line 30: Can we add another test case here:
- form is invalid
- we make a change to an input field
- we assert that the error message is removed

packages/react/src/components/button/index.tsx
```diff
@@ -20,18 +21,30 @@ export const Button: React.FC<Props> = ({
   variant = "primary",
   testId,
   icon,
+  disabled,
 }) => {
+  const handleClick: MouseEventHandler<HTMLButtonElement> = (e) => {
```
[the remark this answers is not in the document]
kaspermroz: you are right it's unnecessary :D will drop it

ikovic (approved):
Looks good, let's merge this 😃

PR #49, opened 2023-02-08
[SID-680] Form onSuccess props

kaspermroz:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-680)

There is a common requirement for a callback prop fired when the authentication via `<Form />` component is successful

## Checklist

### DIFF
--- a/packages/react/src/components/form/form.test.tsx
+++ b/packages/react/src/components/form/form.test.tsx
@@ -75,6 +75,23 @@ describe("#Form", () => {
   test("should show the success state on successful login", async () => {
     const logInMock = vi.fn(() => Promise.resolve(TEST_USER));
     const user = userEvent.setup();
+
+    render(
+      <TestSlashIDProvider sdkState="ready" logIn={logInMock}>
+        <Form />
+      </TestSlashIDProvider>
+    );
+
+    user.click(screen.getByTestId("sid-form-initial-submit-button"));
+
+    await expect(
+      screen.findByTestId("sid-form-success-state")
+    ).resolves.toBeInTheDocument();
+  });
+
+  test("should call the onSuccess callback if provided on a successful login", async () => {
+    const logInMock = vi.fn(() => Promise.resolve(TEST_USER));
+    const user = userEvent.setup();
     const onSuccess = vi.fn();
 
     render(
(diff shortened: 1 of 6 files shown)


packages/react/src/components/form/form.test.tsx
```diff
@@ -75,10 +75,11 @@ describe("#Form", () => {
   test("should show the success state on successful login", async () => {
     const logInMock = vi.fn(() => Promise.resolve(TEST_USER));
     const user = userEvent.setup();
+    const onSuccess = vi.fn();
 
     render(
       <TestSlashIDProvider sdkState="ready" logIn={logInMock}>
-        <Form />
+        <Form onSuccess={onSuccess} />
```
ikovic, line 82: Would you mind adding an additional test case for this? IMO it adds value to have tests as living documentation. So we can have something like `test("should call the onSuccess callback if provided on a successful login", () => ...)`

packages/react/src/components/form/form.test.tsx
```diff
@@ -75,10 +75,11 @@ describe("#Form", () => {
   test("should show the success state on successful login", async () => {
     const logInMock = vi.fn(() => Promise.resolve(TEST_USER));
     const user = userEvent.setup();
+    const onSuccess = vi.fn();
```
ikovic, line 78: Maybe it would make sense to call the callback with the `user` instance that gets created on login success? It could make this more flexible.

packages/react/src/components/form/form.tsx
```diff
@@ -11,9 +11,10 @@ import { useConfiguration } from "../../hooks/use-configuration";
 
 export type Props = {
   className?: string;
+  onSuccess?: () => void;
 };
 
-export const Form: React.FC<Props> = ({ className }) => {
+export const Form: React.FC<Props> = ({ className, onSuccess }) => {
```
ikovic, line 17: Do you think we could pass the `onSuccess` callback to `flowState` instead? This is a matter of flow control, rather than UI rendering. In that case `onSuccess` could be called internally, as soon as login is done? If a customer wants to perform redirect on success, that would happen without `Success` being rendered briefly.

ikovic: Also we'd have a guarantee that the callback was called only once. If we involve `useEffect` it gets tricky :)

ikovic:
@kaspermroz One thing I forgot to bring up is that we also need to document new features - that's done in `ng-evangelion`. You can find the reference docs for `<Form>` there and add the new prop.

ikovic (approved):
Perfect, thank you!

PR #51, opened 2023-02-13
[SID-727] Show divider only when both OIDC and non-OIDC factors are present

kaspermroz:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-727)

We want to make sure that we show `or` divider only when both OIDC and non-OIDC factors are present.

## Checklist

### DIFF
--- a/packages/react/src/components/form/initial.tsx
+++ b/packages/react/src/components/form/initial.tsx
@@ -12,10 +12,12 @@ import {
   FactorOIDC,
   Handle,
   HandleType,
-  isFactorOidc,
-  hasOidcAndNonOidcFactors,
   Validator,
 } from "../../domain/types";
+import {
+  isFactorOidc,
+  hasOidcAndNonOidcFactors,
+} from "../../domain/handles";
 import { Logo as TLogo } from "../../context/config-context";
 import { Flag, GB_FLAG, Input, PhoneInput } from "../input";
 import { TextConfigKey } from "../text/constants";
--- a/packages/react/src/domain/types.ts
+++ b/packages/react/src/domain/types.ts
@@ -14,14 +14,6 @@ export interface LoginOptions {
 
 export type FactorOIDC = Extract<Factor, { method: "oidc" }>;
 
-export function isFactorOidc(factor: Factor): factor is FactorOIDC {
-  return factor.method === "oidc";
-}
-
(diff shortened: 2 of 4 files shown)


packages/react/src/components/form/form.test.tsx
```diff
@@ -25,6 +26,32 @@ describe("#Form", () => {
     );
 
     expect(screen.getByTestId("sid-form-initial-state")).toBeInTheDocument();
+
+    expect(screen.queryByText(TEXT["initial.divider"])).not.toBeInTheDocument();
```
balenio, line 30: @kaspermroz this is **not** my domain, so please put my feedback in context :)
my two cents: imho this test assumes the initial state doesn't have OIDC factors when it shouldn't => I think it'd be more correct to keep this `should render in the initial state` test as-is and add a separate one specific to rendering without the divider where you explicitly don't have OIDC factors.

packages/react/src/components/form/initial.tsx
```diff
@@ -225,6 +225,8 @@ export const Initial: React.FC<Props> = ({ flowState }) => {
     () => factors.filter(isFactorOidc),
     [factors]
   );
+  const hasOidcFactors = Boolean(oidcFactors.length);
```
ikovic, line 228: What if the factors only include the OIDC factors? Eg if we remove all other methods, would we still have this divider displayed?

packages/react/src/domain/types.ts
```diff
@@ -18,6 +18,10 @@ export function isFactorOidc(factor: Factor): factor is FactorOIDC {
   return factor.method === "oidc";
 }
 
+export function hasOidcAndNonOidcFactors(factors: Factor[]): boolean {
```
ikovic, line 21: This is a bit pedantic, not a blocker - a good way of documenting this function would be writing a couple of unit tests that show what the expected behavior is. I would write a test with three cases, the same way you did for the component:
- returns true if both oidc and non oidc are present
- returns false if no oidc factors
- returns false if no non-oidc factors

PR is approved anyway, up to you if you want to add this!

ikovic (approved):
Thank you, looks good! Left a comment related to a unit test, but it's not a blocker. Up to you if you'll merge it as it is or add that test too.

ikovic (approved):
Thanks for the cleanup!

PR #52, opened 2023-02-13
[SID-705] Store last user ID based on configuration

kaspermroz:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-705)

We want to add possibility to store the last user identifier in form (handle type + value). This behaviour should be optional and disabled by default.

## Checklist

### DIFF
--- a/.changeset/early-radios-glow.md
+++ b/.changeset/early-radios-glow.md
@@ -1,0 +1,5 @@
+---
+"@slashid/react": minor
+---
+
+Add new config property `storeLastHandle` (boolean, defaults to `false`) to enable Form autofilling with previously used handle stored in localStorage.


packages/react/src/components/form/initial.tsx
```diff
@@ -139,10 +141,18 @@ const HandleForm: React.FC<HandleFormProps> = ({
           placeholder={text["initial.handle.phone.placeholder"]}
           value={values[handleType] ?? ""}
           flag={flag}
-          onChange={registerField(handleType, (value) => {
-            if (!isValidPhoneNumber(value)) {
-              return { message: text["validationError.phoneNumber"] };
-            }
+          onChange={registerField(handleType, {
+            defaultValue:
+              shouldStoreLastID &&
```
ikovic, line 146: This is quite a complex logical expression. How do you feel about hiding this behind a dedicated function exposed from `useLastIdentifier`? Something like:
```
const { findPreviousHandle } = useLastIdentifier()

// line 145
defaultValue: findPreviousHandle(handleType, defaultValue?)
```

Then we can write tests for this and make sure it behaves exactly as we want it to

packages/react/src/components/form/initial.tsx
```diff
@@ -157,18 +167,42 @@ const HandleForm: React.FC<HandleFormProps> = ({
         label={text["initial.handle.email"]}
         placeholder={text["initial.handle.phone.email"]}
         value={values[handleType] ?? ""}
-        onChange={registerField(handleType, (value) => {
-          if (!isValidEmail(value)) {
-            return { message: text["validationError.email"] };
-          }
+        onChange={registerField(handleType, {
+          defaultValue:
```
ikovic, line 171: This one kind of confirms we should think about the approach proposed in the previous comment 🤓

packages/react/src/components/form/initial.tsx
```diff
@@ -157,18 +167,42 @@ const HandleForm: React.FC<HandleFormProps> = ({
...
+    values,
+    lastIdentifier.handleType,
+    lastIdentifier.value,
+    shouldStoreLastID,
+  ]);
 
   const onSubmit: React.FormEventHandler<HTMLFormElement> = (e) => {
     e.preventDefault();
 
+    if (shouldStoreLastID) {
```
ikovic, line 199: Should this be driven from the UI? Most of the business logic should be implemented either in the `flowState` or the SlashId context.

ikovic (commented):
@kaspermroz one thing I failed to realize earlier - we're still mixing terminology, "identifier" is an obsolete term since we use "handle" instead. Can you please rename this to be something like `useLastHandle`?

packages/react/src/hooks/use-last-handle.ts
```diff
@@ -0,0 +1,59 @@
...
+import { useConfiguration } from "./use-configuration";
+import { useSlashID } from "./use-slash-id";
+
+const STORAGE_LAST_HANDLE_KEY = "@slashid/LAST_HANDLE";
+
+type UseLastHandle = () => {
+  lastHandle: Handle | undefined;
+};
+
+type SuccessEvent = {
```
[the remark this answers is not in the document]
ikovic: @kaspermroz I'll open a ticket for improving this on the SDK side - the way I designed the event types does not really help with the developer experience. Let's leave this as it is for now.

packages/react/src/hooks/use-last-handle.ts
```diff
@@ -0,0 +1,59 @@
...
+  }, []);
+
+  useEffect(() => {
+    if (!storeLastHandle || sdkState !== "loaded" || sid === undefined) {
+      return;
+    }
+
+    if (!subscribed.current) {
+      subscribed.current = true;
+      // @ts-expect-error
```
ikovic, line 44: When adding TS annotations like these, it's a good practice to write a comment about the expected error, something like:

```
// @ts-expect-error TODO core SDK does not export the correct event handler type
```

ikovic: This way it's easier for us to come back and fix this later.

ikovic (approved):
🚢

PR #62, opened 2023-03-10
[SID-767] <MFA /> component

kaspermroz:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-767)

We want to provide Multi Factor Authentication flow to our customers. Please see the [design document](https://docs.google.com/document/d/1gf_hoxI4jIjzL33cMB7uo4EDhnL3fvw7nISOjKmU6Tk/edit#heading=h.1u1y7udks3et) for mode details.

## Checklist

### DIFF
--- a/.changeset/kind-schools-smell.md
+++ b/.changeset/kind-schools-smell.md
@@ -1,0 +1,5 @@
+---
+"@slashid/react": minor
+---
+
+Create <MFA /> component for immediate Multi-Factor Authentication


packages/react/src/components/form/useFlowState.ts
```diff
@@ -16,8 +16,13 @@ export function useFlowState(opts: CreateFlowOptions = {}) {
   }, []);
 
   useEffect(() => {
-    flowRef.current.setLogIn(logIn);
-  }, [logIn]);
+    if (user) {
```
ikovic, line 19: Maybe we could have two refs to avoid this?

ikovic:
@kaspermroz this looks good to me! Can we give a try at creating a `MFAFlow` component (or however we want to name it) that would run the MFA immediately? We might have to make some changes to the `Form` given that the `success` state should not display at all and instead we would render the next MFA step. We could probably control that behavior using `onSuccess` callback.

ikovic (approved):

PR #64, opened 2023-03-17
[SID-818] <LoggedIn> API extension

kaspermroz:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-818)

With exposed MFA functionality, we want to extend `<LoggedIn>` component API. Our customers should be able to specify:
 - a list of authentication factor methods that user has to be logged in
 - a callback function: `(userAuthenticationMethods: FactorMethods) => boolean`

## Checklist

### DIFF
--- a/.changeset/rare-readers-talk.md
+++ b/.changeset/rare-readers-talk.md
@@ -1,0 +1,5 @@
+---
+"@slashid/react": minor
+---
+
+Extend <LoggedIn> API - add `withFactorMethods` prop to specify required authentication factor methods.


packages/react/src/components/logged-in/index.tsx
```diff
@@ -1,16 +1,52 @@
-import { useSlashID } from "../../main";
+import type { Factor, User } from "@slashid/slashid";
+import { useMemo } from "react";
+import { useSlashID } from "../../hooks/use-slash-id";
+
+type FactorMethods = Factor["method"][];
```
ikovic, line 5: You can import this from the core SDK:
```
import type { FactorMethod } from "@slashid/slashid"
```

packages/react/src/components/logged-in/index.tsx
```diff
@@ -1,16 +1,52 @@
...
 
 type Props = {
   children: React.ReactNode;
+  withFactorMethods?:
+    | FactorMethods
+    | ((userAuthenticationMethods: FactorMethods) => boolean);
```
ikovic, line 11: Nice, I like how this enforces mutual exclusivity!

packages/react/src/components/logged-in/logged-in.test.tsx
```diff
@@ -30,4 +30,52 @@ describe("LoggedIn", () => {
     );
     expect(screen.queryByText("Test")).not.toBeInTheDocument();
   });
+
+  test("should render children when user is authenticated with correct method", () => {
+    render(
+      <TestSlashIDProvider user={TEST_USER}>
+        <LoggedIn withFactorMethods={["email_link"]}>
```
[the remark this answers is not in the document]
ikovic: What do you suggest?

packages/react/src/context/test-slash-id-provider.tsx
```diff
@@ -10,11 +10,46 @@ type TestProviderProps = Partial<ISlashIDContext> & {
...
+
+export const TEST_TOKEN = [
+  TEST_TOKEN_HEADER,
+  btoa(JSON.stringify(TEST_TOKEN_PAYLOAD_DECODED)),
+  TEST_TOKEN_SIGNATURE,
+].join(".");
 
 export const TEST_USER = new User(TEST_TOKEN);
 
+export function NewTestUserWithAuthMethods(authMethods: FactorMethod[]): User {
```
ikovic, line 38: This is a factory function so by convention it should be named with `create*` prefix. For example:
```
export function createTestUser({authMethods: FactorMethod[]}): User // ...
```

ikovic: Also it would be useful to document that this will not generate a valid token, just one that can be decoded and used in the SDK for testing.

ikovic: I quite like the idea btw 👏

ikovic (approved):

ikovic:
Looks really good to me, please add to the existing docs and that's it :)

PR #66, opened 2023-03-22
[SID-825] Step-Up Authentication

kaspermroz:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-825)

We want to expose an explicit Step-Up Authentication experience to our customers.

## Checklist

### DIFF
--- a/.changeset/many-hotels-shop.md
+++ b/.changeset/many-hotels-shop.md
@@ -1,0 +1,5 @@
+---
+"@slashid/react": minor
+---
+
+Create <MultiFactorAuth> and <StepUpAuth> components


packages/demo-form/src/pages/FormPage/form-page.tsx
```diff
@@ -0,0 +1,21 @@
+import type { Factor } from "@slashid/slashid";
+import { Form, ConfigurationProvider } from "@slashid/react";
+import styles from "../../styles/Page.module.css";
```
ikovic, line 3: Importing CSS modules not collocated with the component is an indicator of a possibly missing component. Do you think it would make sense to create a `PageLayout` component that would look like this:

```
export function FormPage({containerClassName}) {
  return (
    <main className={styles.page}>
      <section className={styles.section}>
        <div className={containerClassName}>
          {children}
        </div>
      </section>
    </main>
  );
}
```

So then the FormPage would be something like:

```
<PageLayout>
  <ConfigurationProvider factors={factors}>
    <Form />
  </ConfigurationProvider>
</PageLayout>
```

ikovic (commented):
@kaspermroz overall it looks great, I'm quite happy with this. One thing I would add after refactoring the page component would be a heading and a short paragraph that explains what the components are supposed to do. Do you think it makes sense to have that?

packages/react/src/components/multi-factor-auth/multi-factor-auth.tsx
```diff
@@ -1,20 +1,32 @@
...
 };
 
-export function MFA({ factors, text }: MFAProps) {
+/**
+ * First-class Multi-Factor Authentication experience.
+ * This component can be used instead of regular `<Form />`.
+ * After successful authentication with one of the factors defined in a `<ConfigurationProvider>`,
+ * the user will _immediately_ be prompted with second step - one of the `factors`.
+ */
+export function MultiFactorAuth({
```
ikovic, line 20: `Form` should receive an extended set of props in order for developers to be able to override the default values set in the `ConfigurationProvider`. These would be:
- `factors`
- `text`

These values would be optional and if set merged with the `ConfigurationProvider` values.

Then the MultiFactorAuth component could receive an array of steps, each of them having a config entry consisting of the above values.

ikovic: We can override `useConfiguration` to apply overrides to the config provider values.

packages/react/src/components/form/form.tsx
```diff
@@ -33,19 +44,28 @@ export const Form: React.FC<Props> = ({ className, onSuccess }) => {
...
-      )}
-      {flowState.status === "authenticating" && (
-        <FormProvider>
-          <Authenticating flowState={flowState} />
-        </FormProvider>
-      )}
-      {flowState.status === "error" && <Error flowState={flowState} />}
-      {flowState.status === "success" && <Success flowState={flowState} />}
-      <Footer />
+      <ConfigurationProvider
```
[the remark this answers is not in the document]
ikovic: I see, problem is that child components access the context directly, so we cannot override them locally. What you did here makes sense, maybe we could extract it to a component like `FormConfigProvider` which would do the following:
- check if there are any config overrides
- if not, just render `{children}`
- if there are overrides, consume `useConfiguration`, merge the config with the overrides and render another `ConfigurationProvider` with the merge config on top of `{children}`

That would work mostly the same as it is right now but with some separation of concerns, so `Form` is a bit easier to read.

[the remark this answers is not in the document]
ikovic: `ConfigurationOverrides` might be a better name for that extracted component, what do you think?

ikovic (approved):
Nice!

PR #68, opened 2023-03-30
[SID-826] Setup E2E tests

kaspermroz:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-826)

As we release new features and components, we want to make sure that everything is working as expected. While unit tests do cover crucial bits of the logic, we also want to have a shared E2E test suite, which we can run against all of our demo apps in this repository. The additional benefit of this, on top of having actual use cases tested, is that this way we make sure our React SDK works fully with different React frameworks (Next.js, Remix, etc.).

## Checklist


ikovic (approved):
Approved - good work! I tested these locally and it only took ~12 seconds to spin up and run everything.

PR #69, opened 2023-04-12
[SID-850] Exclude font from the bundle

kaspermroz:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-850)

We want to exclude Inter font from our CSS bundle - with the font the file is ~4MiB, without ~22KiB

Screenshot 2023-04-12 at 10 16 35

## Checklist

### DIFF
--- a/packages/react/index.html
+++ b/packages/react/index.html
@@ -4,6 +4,10 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Vite + React + TS</title>
+    <link
+      rel="stylesheet"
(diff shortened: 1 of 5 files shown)


packages/react/index.html
```diff
@@ -4,6 +4,10 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Vite + React + TS</title>
+    <link
+      rel="stylesheet"
+      href="https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600&display=swap"
```
kaspermroz, line 9: This is for development only

PR #72, opened 2023-04-14
Adjust Remix app

kaspermroz:
## Description

Minor Remix app align with the docs update.

## Checklist

### DIFF
--- a/apps/react-remix/app/slash-id.tsx
+++ b/apps/react-remix/app/slash-id.tsx
@@ -12,7 +12,7 @@ export const SlashID = ({ children }: Props) => {
   return (
     <SlashIDProvider
       baseApiUrl="https://api.sandbox.slashid.com"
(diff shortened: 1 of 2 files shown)


apps/react-remix/app/slash-id.tsx
```diff
@@ -12,7 +12,7 @@ export const SlashID = ({ children }: Props) => {
   return (
     <SlashIDProvider
       baseApiUrl="https://api.sandbox.slashid.com"
-      oid={process.env.SLASHID_ORGANIZATION_ID ?? ""}
+      oid="ORG_ID"
```
kaspermroz, line 15: I have no idea how this worked previously, as Remix does not expose env vars to the client (unlike Next.js)

ikovic: It did not work :)

ikovic (approved):

PR #83, opened 2023-06-13
[SID-948] Allow passing in an external token to the SDK

ikovic:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-948)

Allow passing in an external token to the SDK (some clients might store the token somewhere else).

## Checklist

### DIFF
--- a/packages/react/src/context/slash-id-context.tsx
+++ b/packages/react/src/context/slash-id-context.tsx
@@ -15,6 +15,7 @@ export type StorageOption = "memory" | "localStorage";
 
 export interface SlashIDProviderProps {
   oid: string;
+  token?: string;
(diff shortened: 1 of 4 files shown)


packages/react/src/context/slash-id-context.tsx
```diff
@@ -15,6 +15,7 @@ export type StorageOption = "memory" | "localStorage";
 
 export interface SlashIDProviderProps {
   oid: string;
+  token?: string;
```
kaspermroz, line 18: please adjust the name to `initialToken` :)

ikovic: Done already!

kaspermroz (approved):

PR #89, opened 2023-07-13
[SID-1094] Add <SlashIDLoaded> guard component

jake-slashid:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1094)

## Context
When using the React SDK it is often assumed in how the SDK has been implemented that the core SDK is available and ready to use. While this may be true depending on the state of consumers application, this often is not true during the initial bootstrap stage of a React app.

For example when implementing `<LoggedOut>`:
```jsx
<LoggedOut>
  Please sign in!
</LoggedOut>
```
"Please sign in!" will be momentarily displayed while the SDK is being bootstrapped.

### Feature
- Add `@faker-js/faker`: I'm using this to help avoid false positives in testing by randomly generating data in tests.
- Add `<SlashIDLoaded>`
- Implement `<SlashIDLoaded>` in `<LoggedOut>`
- I'm forgoing any other changes for now
    - Since most other assertions are positive assertions `if user === truthy` - although implicitly - the falsy SDK ready states are already covered
    - In future I'd like these checks to be more explicit so we don't get bitten by assuming one property (`user`) is a reasonable proxy for another (`sdkState`), and then later it changes in behaviour.

### Documentation
- See the reference documentation in [this companion PR](https://github.com/slashid/ng-evangelion/pull/843)

## Checklist


jake-slashid:
TODO: documentation, working on it

packages/react/src/components/test-utils.ts
```diff
@@ -16,3 +17,7 @@ export const inputPhone = (
   const input = screen.getByPlaceholderText(inputPlaceholder);
   fireEvent.change(input, { target: { value } });
 };
+
+export const sdkNotReadyStates = Object
+  .values(SDKState)
+  .filter((state): state is Exclude<SDKState, SDKState.Ready> => state != SDKState.Ready)
```
jake-slashid, line 23: This is a practical example of where `enum` is helpful, thanks to this we avoided duplicating the states into an array literal that is destined to go stale.

The anti pattern:
```ts
type SDKState = "a" | "b" | "ready"

const sdkNotReadyStates: Exclude<SDKState, "ready"> = [
  "a",
  "b"
]

// looks ok so far, except above is exclusive assertion and not an inclusive one
// it does not assert that all members of SDKState must be present, but they are
// at the time of writing

// add another sdk state
type SDKState = "a" | "b" | "c" | "ready"

// 💥 our tests miss an edge case, and there are no type errors to warn us.
```

Thanks to using an enum, we actually reference the source of truth so we always have an up-to-date list of states.

packages/react/src/components/loaded/loaded.tsx
```diff
@@ -0,0 +1,32 @@
...
+    const { sdkState } = useSlashID()
+
+    if (sdkState !== SDKState.Ready) {
+        if (!fallback) return null
+
+        return (
+            <>
+                {fallback}
+            </>
+        )
```
paulo-raca, line 24: Just curious, why do you need the <>?

Wouldn't `return fallback` work?

packages/react/src/components/logged-out/logged-out.test.tsx
```diff
@@ -5,13 +5,15 @@ import {
   TestSlashIDProvider,
   TEST_USER,
 } from "../../context/test-slash-id-provider";
+import { SDKState } from "../../context/slash-id-context";
+import { sdkNotReadyStates } from "../test-utils";
 
 const TestComponent = () => <h1>Test</h1>;
```
paulo-raca, line 11: nit: Maybe use `faker.lorem.sentence()` here too

packages/react/src/components/loaded/loaded.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+      render(
+        <TestSlashIDProvider sdkState={state}>
+          <SlashIDLoaded>
+            <Mock text={text} />
+          </SlashIDLoaded>
+        </TestSlashIDProvider>
+      );
+      expect(screen.queryByText(text)).not.toBeInTheDocument();
+    });
+  }
```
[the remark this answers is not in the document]
jake-slashid: 🤔 I'm not sure how this ended up here, might have been a mistake when I was copying the test to `logged-out.test.tsx`

packages/react/src/components/loaded/loaded.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  for (const state of sdkNotReadyStates) {
+    test(`should render fallback when sdk state is '${state}'`, () => {
+      const text = faker.lorem.sentence()
+      const fallback = faker.lorem.sentence()
+
+      render(
+        <TestSlashIDProvider sdkState={state}>
+          <SlashIDLoaded
+            fallback={<Mock text={fallback} />}
```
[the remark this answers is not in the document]
jake-slashid: No not at all, passing JSX to a prop is standard and encouraged. This is one of the key strengths of JSX in general compared to other templating languages.

The React builtin component `<Suspense>` does exactly this:
https://react.dev/reference/react/Suspense

What you're suggesting isn't necessarily bad but I feel like the trade off might not be worth it because it increases our API surface, whereas this pattern is well understood by React developers and it keeps our API surface smaller.

packages/react/src/components/loaded/loaded.tsx
jake-slashid: I definitely agree that it's verbose and pretty ugly.

Technically speaking yes we can `return fallback` because we do a falsy check on line 18, so `fallback` must be a truthy `React.ReactNode` - so we can return it, surely?

In practice this doesn't work because `React.ReactNode` is not perfectly compatible with the `React.FC<...>` expected return type.

```
Type '({ fallback, children }: Props) => string | number | true | ReactFragment | Element | null' is not assignable to type 'FC<Props>'.
  Type 'string | number | true | ReactFragment | Element | null' is not assignable to type 'ReactElement<any, any> | null'.
    Type 'string' is not assignable to type 'ReactElement<any, any>'.
```

By wrapping `fallback` [in a fragment](https://react.dev/reference/react/Fragment) we comply with the return type and cover a bunch of edge cases.

packages/react/src/components/logged-out/logged-out.test.tsx
```diff
@@ -22,12 +24,25 @@ describe("LoggedOut", () => {
...
         <LoggedOut>
           <TestComponent />
         </LoggedOut>
       </TestSlashIDProvider>
     );
     expect(screen.queryByText("Test")).not.toBeInTheDocument();
   });
+
+  for (const state of sdkNotReadyStates) {
+    test(`should not render children when sdk state is ${state}`, () => {
```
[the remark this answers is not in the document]
jake-slashid: Preventing this behaviour is actually one of the reasons this piece of work was needed.

Check out the context section in the PR description. 

Also check out the documentation for this component:
[<LoggedOut> | /id Documentation](https://developer.slashid.dev/docs/react-sdk/reference/react-sdk-reference-loggedout)

It says the content shouldn't be shown until post-initialisation, but in practice this isn't how it works. Showing that content pre-init is a bug.

packages/react/src/components/logged-out/logged-out.test.tsx
jake-slashid: Good idea, let me add it!

packages/react/src/components/loaded/loaded.tsx
kaspermroz: @jake-slashid React.FC is unnecessary (and I think it's still discouraged by the community?) for the most part, as TS can infer the type quite well. I know that we're using it quite heavily, but maybe it's time to stop doing it for the new components? cc @ikovic

packages/react/src/context/slash-id-context.tsx
```diff
@@ -23,12 +23,13 @@ export interface SlashIDProviderProps {
...
-  | "retrievingToken"
-  | "authenticating"
-  | "ready";
+export enum SDKState {
+  Initial = "initial",
+  Loaded = "loaded",
+  RetrievingToken = "retrievingToken",
+  Authenticating = "authenticating",
+  Ready = "ready"
+}
```
kaspermroz, line 32: Please don't use TS enums 💀 We've had tremendous amount of issues with them in the past, es they do not have proper values in the runtime. I don't think it's worth to introduce them here

kaspermroz: I'm also up for discussion if you think otherwise @jake-slashid @ikovic

packages/react/src/components/loaded/loaded.tsx
jake-slashid: Full transparency: I don't have a strong opinion about this, I seen it used in another component and was following that.

packages/react/src/context/slash-id-context.tsx
jake-slashid: Can you explain what exactly the problem was, or give me some reading?

I've used enums for as long as I've used ts, and never experienced any problem. At compile time this just becomes an object.

[post transpilation example](https://www.typescriptlang.org/play?#code/KYOwrgtgBAygIgaRgFwIbOFA3gKClASRAEtljUAbKAXigCJiSzK6AaPKAGQHtUATYHxr0KvAXzYcASsGQAnYsABujAOYAVbgGtQwunNkLlazTpCT8AQTDIAFqDIBjdGr2ob9kE5chVFqDL8AJ56BsF0OAC+QA)

kaspermroz: We had TS errors when trying to use Core SDK in the React SDK, I remember it was connected to the Factors 🤔 You can see how we pretty much removed them from the Core SDK here: https://github.com/slashid/ng-evangelion/pull/610

jake-slashid: 🤔 

Let me try to reproduce something like that

jake-slashid: So here's the outcome of that investigation, everything seems in order to me! The `*.d.ts` file looks sensible.

In the transpiled js whenever it's not necessary to actually reference, enum refs are converted to their literal value:
```ts
// before
if (!sid) {
  return;
}

setState(SDKState.Authenticating); // it's an enum

try {
  const identifier =
    factor.method === "oidc" || handle === undefined
      ? null
      : {
          type: handle.type as unknown as PersonHandleType,
          value: handle.value,
        };

  // @ts-expect-error TODO make the identifier optional
  const user = await sid.id(oid, identifier, factor);
  storeUser(user);
  setState(SDKState.Ready);
  return user;
} catch (e) {
  setState(SDKState.Ready);
  logOut();
  throw e;
}

// after
if (!!se) {
  _("authenticating"); // it's a string now
  try {
    const Oe = Z.method === "oidc" || re === void 0 ? null : {
      type: re.type,
      value: re.value
    }, de = await se.id(f, Oe, Z);
    return U(de), _("ready"), de;
  } catch (Oe) {
    throw _("ready"), J(), Oe;
  }
}
```

For cases where there is a reference, that can't be made literal this happens:
```ts
// main.js
const g9 = Object.values(sv).filter((f) => f != sv.Ready);
export {
  g9 as sdkNotReadyStates
}

// main.d.ts
import { SDKState } from "./context/slash-id-context";
export declare const sdkNotReadyStates: (SDKState.Initial | SDKState.Loaded | SDKState.RetrievingToken | SDKState.Authenticating)[];
```

packages/react/src/components/loaded/loaded.tsx
ikovic: By now I think it's rather clear that `React.FC` is mostly abandoned when compared to the following:
- adding an explicit children field in the props type declaration in case the component really needs children
- adding an explicit return type like JSX.Element

My vote would be to move to that approach as we work on maintaining the existing components and implementing the new ones.

packages/react/src/context/slash-id-context.tsx
ikovic: I do not have an example handy, the issue we faced was as Kasper described - enum values were not present during runtime. It was a specific set of circumstances though - I believe some of these enums were generated from the API spec, then imported to the core SDK and when referencing them from projects that install the core SDK their values wouldn't be present during runtime.

Is there anything we cannot do with an union specifically to complete this task @jake-slashid ?

packages/react/src/components/loaded/loaded.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+interface Props {
+    fallback?: React.ReactNode
+    children: React.ReactNode
+}
+
+/**
+ * Conditional rendering helper
+ * 
+ * Acts as a guard for SlashID core SDK dependent operations
```
ikovic, line 12:
```suggestion
 * Acts as a guard for SlashID core SDK dependent operations. Can optionally render a fallback component while the SDK is loading.
```

packages/react/src/components/loaded/loaded.tsx
```diff
@@ -0,0 +1,32 @@
...
+ */
+export const SlashIDLoaded: React.FC<Props>  = ({ fallback, children }) => {
+    const { sdkState } = useSlashID()
+
+    if (sdkState !== SDKState.Ready) {
```
ikovic, line 17: Not including `authenticating` here might have side effects - whenever `logIn` is executed, state switches from `ready` to `authenticating`.

When I was working on that initially I figured it would make sense but now I'm having second thoughts as I cannot even think of a use case for that - authentication states are handled inside the `<Form>`.

So IMO the solution would be to actually remove the `authenticating` state from the context given that it does not really correspond to the lifecycle of the SDK. What do you think @jake-slashid? Of course we should look for any side effects of that if we do remove it.

packages/react/src/context/slash-id-context.tsx
jake-slashid: @ikovic this line of code depends on it, but it's not super important.

https://github.com/slashid/javascript/pull/89/files#diff-3b087a4ccb39999a6946731aae6334d8e3ed07d09a106ffd27fc74763c567ac0R21-R23

To remove the enum, this:
```ts
export const sdkNotReadyStates = Object
  .values(SDKState)
  .filter((state): state is Exclude<SDKState, SDKState.Ready> => state != SDKState.Ready)
```

Becomes this:
```ts
export const sdkNotReadyStates: SDKState[] = [
  "initial",
  "loaded",
  "retrievingToken",
  "authenticating",
  "ready"
]
```

packages/react/src/components/loaded/index.tsx
```diff
@@ -0,0 +1 @@
+export { SlashIDLoaded } from './loaded'
```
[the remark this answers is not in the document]
jake-slashid: I'm happy to change it and drop the barrel file, again not taking a strong stance here this time - just trying to follow what I perceived as convention and get my PR merged 😅 

`file-upload` does this pattern, where as `dropdown` and `divider` don't so I wasn't sure.

packages/react/src/context/slash-id-context.tsx
jake-slashid: The trade-off is that `sdkNotReadyStates` might go stale in future, compiler doesn't help us here.

packages/react/src/components/loaded/loaded.tsx
jake-slashid: Thanks for your input! Here's the PR where it gets soft-deprecated: https://github.com/facebook/create-react-app/pull/8177

I'll remove React.FC and simplify the component. 🙌

packages/react/src/components/loaded/loaded.tsx
ikovic: Agreed then! 👍 Let's remove it 😄

packages/react/src/components/loaded/index.tsx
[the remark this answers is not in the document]
ikovic: Yep some of the components were not written by us - we got them with the KYC SDK, did not refactor anything at the time since the code is currently unused.

packages/react/src/components/loaded/loaded.tsx
jake-slashid: done

packages/react/src/components/test-utils.ts
jake-slashid: Great idea, this satisfies both arguments!

jake-slashid: I moved both `sdkStates`, `SDKState` and `sdkNotReadyStates` into `domain/sdk-state.ts`. I think it's a good idea to keep these colocated.

packages/react/src/context/slash-id-context.tsx
jake-slashid: done

packages/react/src/components/loaded/index.tsx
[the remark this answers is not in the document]
jake-slashid: done!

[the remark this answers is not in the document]
jake-slashid: Thanks for extra context 🙏

packages/react/src/components/loaded/loaded.tsx
jake-slashid: It's gone 🪄 ✨

ikovic (approved):

kaspermroz (approved):

Issue #94, opened 2023-07-19
Create the GDPR storage interface


ikovic:
We need to use a storage interface to fetch and update GDPR consent. When a customer is authenticated using SlashID, we should use the corresponding `User` instance. Otherwise local storage should be used instead.

PR #96, opened 2023-07-20
[Issue-94] Create the GDPR storage interface

ahmed-gz:
## Description

[Github issue](https://github.com/slashid/javascript/issues/94)

Added a new hook `useGdprConsent` that returns the user's GDPR consents and allows to delete and update them.

## Checklist

### DIFF
--- a/packages/react/src/hooks/use-gdpr-consent.ts
+++ b/packages/react/src/hooks/use-gdpr-consent.ts
@@ -7,7 +7,7 @@ export const STORAGE_GDPR_CONSENT_KEY = "@slashid/GDPR_CONSENT";
 
 type UseGdprConsent = () => {
   consents: GDPRConsent[];
-  updateGdprConsent: (consent: GDPRConsentLevel[]) => Promise<void>;
+  updateGdprConsent: (consentLevels: GDPRConsentLevel[]) => Promise<void>;
 };
 
 const getStoredConsents = (): GDPRConsent[] => {
@@ -16,7 +16,6 @@ const getStoredConsents = (): GDPRConsent[] => {
   }
 
   const storedConsents = window.localStorage.getItem(STORAGE_GDPR_CONSENT_KEY);
-
   if (storedConsents === null) {
     return [];
   }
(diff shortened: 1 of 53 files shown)


kaspermroz (approved):

packages/react/src/hooks/use-gdpr-consent.ts
```diff
@@ -0,0 +1,128 @@
...
+  getConsentLevels: () => Promise<GDPRConsent[]>;
+  setConsentLevels: (
+    consentLevels: GDPRConsentLevel[]
+  ) => Promise<GDPRConsent[]>;
+  deleteConsentLevels: () => Promise<void>;
+}
+
+export const createLocalGDPRConsentStorage = (): GDPRConsentStorage => ({
+  getConsentLevels: async () => {
+    if (!isBrowser()) {
```
ikovic, line 18: This can be the responsibility of the parent scope - we can check it there and only create an instance if we're in a browser, no need to have local checks then.

packages/react/src/hooks/use-gdpr-consent.ts
```diff
@@ -0,0 +1,128 @@
...
+
+export const useGdprConsent: UseGdprConsent = () => {
+  const { user, sid } = useSlashID();
+  const [consents, setConsents] = useState<GDPRConsent[]>([]);
+
+  const storage = useMemo(() => {
+    return user
+      ? createApiGDPRConsentStorage(user)
+      : createLocalGDPRConsentStorage();
+  }, [user]);
```
ikovic, line 98: IMO we should check for the browser here too. Eg if there is no user, we should also check if we're in a browser before trying to use local storage.

ikovic:
@ahmed-gz IMO the changes look good, could you please write tests for these cases:
- anonymous user, visits for the first time
- anonymous user, return visit
- anonymous user converts to an authenticated user
- anonymous user, return visit

I believe you should be able to write tests using the testing library and client mocking in a similar way we had so far without having to resort to playwright - please check the existing tests for examples.

PR #98, opened 2023-07-25
[SID-1126] Add isLoading to SlashIDContext

jake-slashid:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1126)

There is no way to check via the `useSlashID()` hook if the SDK is currently "loading" without using internal `sdkState`. We have `<SlashIDLoaded />` but this only helps for composition in JSX, it doesn't satisfy the use case of early-return style conditional rendering.

See YouTrack for more detailed context.

## Checklist


jake-slashid:
Closing in favour of https://github.com/slashid/javascript/pull/89

This change was added there

PR #105, opened 2023-07-27
[SID-1137] feat: add useOrganization and <OrganizationSwitcher />

jake-slashid:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1137)

## Changes
- Add `useOrganizations()` hook
- Add `<OrganizationSwitcher />` component
- [Documentation](https://github.com/slashid/ng-evangelion/pull/930)

## Checklist


jake-slashid:
For simplicity I'm considering
```tsx
<Dropdown
  disabled={isLoading}
  ...
/>
```

packages/react/package.json
```diff
@@ -55,7 +55,7 @@
   },
   "devDependencies": {
     "@faker-js/faker": "^8.0.2",
-    "@slashid/slashid": "^3.9.0",
+    "@slashid/slashid": "link:/Users/jake/dev/ng-evangelion/web/sdk",
```
ikovic, line 58: Leaving a note here so we don't forget replacing this with the actual package :)

packages/react/src/components/organization-switcher/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { useOrganizations } from "../../hooks/use-organizations"
+import { sprinkles } from "../../theme/sprinkles.css";
+import { Dropdown } from "../dropdown"
+import { Spinner } from "../spinner/spinner";
+
+export function OrganizationSwitcher() {
+  const { text } = useConfiguration();
+  const { organizations, currentOrganization, switchOrganization, isLoading } = useOrganizations()
+
+  if (isLoading || !currentOrganization) return <Spinner /> // need a better loading state
```
ikovic, line 11: This is the first time we offer a component that has to load something first - maybe it needs an optional loading fallback prop? Alternatively we could disable the input while it's loading 🤔

packages/react/src/context/slash-id-context.tsx
```diff
@@ -16,6 +16,7 @@ export type StorageOption = "memory" | "localStorage";
 
 export interface SlashIDProviderProps {
   oid: string;
+  onOidChange?: (oid: string) => void | Promise<void>
```
ikovic, line 19: Are we sure we need this? Maybe a direct callback could be registered on the hook itself?

packages/react/src/components/organization-switcher/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+  return (
+    <Dropdown
+      defaultValue={currentOrganization.id}
+      className={sprinkles({ marginBottom: "3", marginTop: "5" })}
```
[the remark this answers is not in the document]
jake-slashid: Good idea 👍

packages/react/src/context/slash-id-context.tsx
jake-slashid: It's not really needed to be honest, but equally neither is the reactive behaviour for the `oid` prop using `useEffect`.

I'm happy to treat `oid` as `initialOid` and be done with it, we can rename later.

packages/react/src/components/organization-switcher/index.tsx
ikovic: Sounds great, thank you!

packages/react/src/hooks/use-organizations.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  const currentOrganization = useMemo(() => {
+    if (!user) return null
+
+    return organizations.find(org => org.id === user.oid) ?? null
+  }, [organizations, user, user?.oid])
+
+  useEffect(() => {
+    if (!user) return
+    if (organizations.length) return
```
kaspermroz, line 24: You could combine these I think

paulo-raca (commented):
LGTM!

packages/react/src/hooks/use-organizations.ts
jake-slashid: done, thanks! 🙌

packages/react/src/components/organization-switcher/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+const className = `sid-organization-switcher ${sprinkles({ marginBottom: "3", marginTop: "5" })}`
+
+const DefaultFallback = () => (
+  <Dropdown
+    defaultValue={undefined}
+    disabled={true}
+    className={className}
+    items={[]}
+    onChange={() => {}}
+    label={""}
```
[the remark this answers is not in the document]
jake-slashid: I added the label

jake-slashid:
Documentation: https://github.com/slashid/ng-evangelion/pull/930

packages/react/src/components/organization-switcher/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+        styles.organizationSwitcher,
+        "sid-organization-switcher"
+      )}
+    >
+      {children}
+    </div>
+  );
+};
+
+export const OrganizationSwitcher = ({
```
ikovic, line 53: Some JS doc would come in handy here 😃

ikovic (approved):
Just a minor comment about JS doc, otherwise looks great!

packages/react/src/components/organization-switcher/index.tsx
jake-slashid: done

Issue #108, opened 2023-07-31
Implement the Modal component


ikovic:
This component would act as a wrapper to the GDPR dialog component (but also be able to receive any child component) in case customers want us to handle layout too. It should receive props that allow controlling the modal open state. It should also implement an overlay that prevents interaction with the background (we can use Radix for this).

PR #110, opened 2023-08-01
[Issue-108] Implement the Modal component

ahmed-gz:
## Description

[Github issue](https://github.com/slashid/javascript/issues/108)

Added a new component `<Dialog>` that can act either as a modal or dialog based on `modal` prop to allow or prevent interaction with outside elements.

## Checklist


packages/react/src/components/dialog/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+  children: ReactNode;
+  trigger: ReactNode;
+  open: boolean;
+  onOpenChange: (open: boolean) => void;
+  icon?: ReactNode;
+  className?: string;
+  modal?: boolean;
+};
+
+export const Dialog = ({
```
ikovic, line 17: @ahmed-gz please add a JS doc style comment here to explain the props

packages/react/src/components/dialog/style.css.ts
```diff
@@ -0,0 +1,58 @@
...
+  animation,
+});
+
+export const wrapper = style({
+  position: "fixed",
+  top: "50%",
+  left: "50%",
+  transform: "translate(-50%, -50%)",
+  width: "90vw",
+  maxWidth: "398px",
```
ikovic, line 24: Can we transition over the height property so the dialog does not jump when the content changes height?

ikovic (approved):

PR #112, opened 2023-08-02
Implement the react query poc

ikovic:
## Description

## Checklist

### DIFF
--- a/packages/react/src/hooks/use-organizations.ts
+++ b/packages/react/src/hooks/use-organizations.ts
@@ -1,0 +1,16 @@
+import { useQuery } from "@tanstack/react-query";
+import { useSlashID } from "./use-slash-id";
+
+export function useOrganizations() {
+  const { isAuthenticated, user } = useSlashID();
+
+  const result = useQuery({
+    queryKey: ["organizations", user?.oid],
+    queryFn: () => {
+      return user?.getOrganizations();
+    },
+    enabled: isAuthenticated,
+  });
+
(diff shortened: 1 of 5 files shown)


packages/react/src/hooks/use-organizations.ts
```diff
@@ -0,0 +1,16 @@
+import { useQuery } from "@tanstack/react-query";
+import { useSlashID } from "./use-slash-id";
+
+export function useOrganizations() {
+  const { isAuthenticated, user } = useSlashID();
+
+  const result = useQuery({
+    queryKey: ["organizations", user?.oid],
```
ikovic, line 8: We could include the person ID here too for other cases to segregate the cache entries.

packages/react/src/hooks/use-organizations.ts
```diff
@@ -0,0 +1,16 @@
...
+    queryFn: () => {
+      return user?.getOrganizations();
+    },
+    enabled: isAuthenticated,
```
ikovic, line 12: The query is in initial mode as long as this flag is false, it will automatically fetch data when `isAuthenticated` changes to `true`.

packages/react/src/hooks/use-organizations.ts
```diff
@@ -0,0 +1,16 @@
...
+  });
+
+  return result;
```
ikovic, line 15: A stable interface is returned - the type signature is inferred correctly

PR #113, opened 2023-08-02
[SID-1146] Set the theme root on the SlashIDProvider level

ikovic:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1146)

Set the theme root on the SlashIDProvider level to make sure all the components inherit the variables set there.
Marked the `theme` prop on the `ConfigurationProvider` as deprecated.

## Checklist

### DIFF
--- a/packages/react/src/components/form/form.tsx
+++ b/packages/react/src/components/form/form.tsx
@@ -5,7 +5,6 @@ import { Initial } from "./initial";
 import { Authenticating } from "./authenticating";
 import { Error } from "./error";
 import { Success } from "./success";
-import { themeClass, darkTheme, autoTheme } from "../../theme/theme.css";
 import * as styles from "./form.css";
 import { Footer } from "./footer";
 import { useConfiguration } from "../../hooks/use-configuration";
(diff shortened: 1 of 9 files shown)


packages/react/src/context/config-context.tsx
```diff
@@ -2,14 +2,17 @@ import { Factor } from "@slashid/slashid";
...
 
-export type Theme = "light" | "dark" | "auto";
 export interface IConfigurationContext {
   text: TextConfig;
   factors: Factor[];
   logo: Logo;
+  /**
+   * @deprecated Set this on the SlashIDProvider instead using the themeProps prop.
+   */
   theme: Theme;
```
jake-slashid, line 16: This feels like a breaking change more than a deprecation, we're no longer referencing it so this property doesn't work anymore.

packages/react/src/components/form/form.tsx
```diff
@@ -28,20 +27,11 @@ export const Form: React.FC<Props> = ({
   text,
 }) => {
   const flowState = useFlowState({ onSuccess });
-  const { theme, showBanner } = useConfiguration();
```
jake-slashid, line 31: I think to truly deprecate this we need to still support this for some time before removing it

I guess something like:
```ts
const { theme } = useConfiguration()
const root = useRootThemeProps()

const t = theme ?? root.theme
const cn = `${className} ${root.className}`
``` 

and then duplicate the `<ThemeRoot>` implementation to be in this component too.

jake-slashid (approved):

PR #116, opened 2023-08-03
[SID-1153] Introduce middleware

jake-slashid:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1153)

We'd like to provide a robust API for mutating the user object or performing side-effects, prior to completion of login.

Some example use cases:
- Adding attributes captured via a custom form to a user bucket
- Switching the user from the root org to another org
- Querying information about the user to hydrate application state

## Changes
- Add `middleware` prop to `<Form />` and `logIn()`.
- Add our first middleware: `defaultOrganization`.
- Remove the `defaultOrganization` prop from `<SlashIdProvider />`
- Remove `<OrganizationProvider />`

### Examples
Given an org selector function:
```ts
const orgSelector = (orgs) => {
  const preferred = orgs.find((org) => org.org_name === "PREFERRED_ORG");
  if (preferred) return preferred.id;

  return "ROOT_OID";
}
```

`<Form>`
```tsx
import { defaultOrganization } from '@slashid/react/middleware'

...

<Form
  middleware={[
    defaultOrganization(orgSelector)
  ]}
/>
```
`logIn()`
```ts
import { useSlashID } from '@slashid/react'
import { defaultOrganization } from '@slashid/react/middleware'

const { logIn } = useSlashID()

const factor = { method: "email_link" }
const handle = { value:"foo@bar.com", type: "email_address" }

logIn({ factor, handle }, {
  middleware: [
    defaultOrganization(orgSelector)
  ]
})
```

## Checklist


packages/react/src/middleware/default-organization.ts
```diff
@@ -0,0 +1,13 @@
+import { OrganizationDetails, User } from "@slashid/slashid";
+import { LoginMiddleware } from "../domain/types";
+
+export const defaultOrganization = (oid: string | ((organizations: OrganizationDetails[]) => string)): LoginMiddleware => async ({ user }) => {
```
ikovic, line 4: This looks good! Maybe consider adding some inline docs in TS Doc format so that pops up in the editor?

packages/react/src/hooks/use-organizations.ts
```diff
@@ -1,19 +1,38 @@
...
-export interface UseOrganizations extends IOrganizationContext {
-  switchOrganization: ({ oid }: { oid: string }) => void;
+export interface UseOrganizations {
+  organizations: OrganizationDetails[]
+  currentOrganization: OrganizationDetails | null
+  switchOrganization: ({ oid }: { oid: string }) => void
+  isLoading: boolean
 }
 
 export const useOrganizations = (): UseOrganizations => {
```
ikovic, line 14: IMO it would be worth the effort to include some TS Docs here.

packages/react/src/context/slash-id-context.tsx
```diff
@@ -239,21 +208,21 @@ export const SlashIDProvider: React.FC<SlashIDProviderProps> = ({
...
-        if (!user || userFromToken.token !== user.token) {
-          setUser(userFromToken);
-        }
-      };
-      slashId.subscribe("idFlowSucceeded", getUserFromEvent);
-
-      return () => slashId.unsubscribe("idFlowSucceeded", getUserFromEvent);
-    }
-  }, [state, user]);
+  // useEffect(() => {
```
ikovic, line 211: Let's try deleting this, we have it in git anyway if it needs to be resurrected

jakewhelan: Was leaving it here to get your thoughts! I will remove it

packages/react/src/hooks/use-organizations.ts
```diff
@@ -1,19 +1,38 @@
...
+      })
+  }, [user])
+
+  const currentOrganization = useMemo(() => {
+    if (!user) return null
+
+    return organizations.find(org => org.id === user.oid) ?? null
+  }, [organizations, user])
+
+  const isLoading = useMemo(() => Boolean(!organizations.length) || !currentOrganization, [organizations, currentOrganization])
```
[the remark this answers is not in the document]
jakewhelan: Oops yes you're right

packages/react/src/middleware/default-organization.ts
jake-slashid: done

Issue #109, opened 2023-07-31
Implement the GDPRConsentDialog component


ikovic:
Implement the `GDPRConsentDialog` component using the already implemented hook as the backing.

PR #118, opened 2023-08-04
[Issue-109] Implement the GDPRConsentDialog component

ahmed-gz:
## Description

[Github issue](https://github.com/slashid/javascript/issues/109)

- Implemented the `<GDPRConsentDialog>` component to manage the GDPR consent levels for the current user using the existing `useGdprConsent` hook.
- Updated the existing `useGdprConsent` hook to allow control over the rendering cycles using a state variable.
- Implemented the `<Accordion>` component built on top of Radix Accordion primitives.
- Implemented the `<Switch>` component built on top of Radix Switch primitives.
- Added icons and new variations to the `<Button>` component.

## Checklist


packages/react/src/components/accordion/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+  content: ReactNode;
+};
+
+type Props = {
+  items: Item[];
+  className?: string;
+  itemClassName?: string;
+};
+
+export const Accordion = ({ items, className, itemClassName }: Props) => (
```
ikovic, line 20: @ahmed-gz this will be for internal use but still it would be useful to have some docs here on how to use it

packages/react/src/components/gdpr-consent-dialog/state.ts
```diff
@@ -0,0 +1,100 @@
+import { GDPRConsent } from "@slashid/slashid";
+
+export const GDPR_CONSENT_LEVELS = [
+  "necessary",
```
ahmed-gz, line 4: Add `none` and filter it out when needed

packages/react/src/components/gdpr-consent-dialog/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+        payload: consents,
+      });
+    }
+  }, [consentState, consents]);
+
+  if (consentState === "initial") {
+    return null;
+  }
+
+  return (
```
ahmed-gz, line 87: Wrap with `<SlashIDLoaded>` here

packages/react/src/components/gdpr-consent-dialog/state.ts
```diff
@@ -0,0 +1,100 @@
...
+  switch (action.type) {
+    case "SET_OPEN":
+      return {
+        ...state,
+        open: action.payload,
+      };
+    case "SET_CONSENT_SETTINGS":
+      return {
+        ...state,
+        consentSettings: getConsentSettings(action.payload),
```
ahmed-gz, line 64: Pull this transformation to the component and memorized it

packages/react/src/components/gdpr-consent-dialog/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+    }
+  }, [open]);
+
+  useEffect(() => {
+    if (consentState === "ready") {
+      dispatch({
+        type: "SET_CONSENT_SETTINGS",
+        payload: consents,
+      });
+    }
```
ahmed-gz, line 80: We should remove `SET_CONSENT_SETTINGS` case in the reducer and use derived state and `useMemo` for `consentSettings` and `open`

ikovic:
> Should this have storybook stories?
> 
> Because it's customer facing I expected to find an example but it's not there
> 
> @ahmed-gz @ikovic what do you think?

It definitely should! @ahmed-gz if there are no outstanding comments on the PR I would suggest working on the stories as the next step

jake-slashid:
Sorry if this ends up being a bit spammy - I'll be posting my thoughts here as I progress through an implementation of the component in my demo app

1. `GDPRConsentDialog` needs to be exported in `main.ts`

ahmed-gz:
> ding comments on

I just need to add some details to some unit tests then I will jump right on it

jake-slashid:
This feedback is mostly for the underlying `Dialog`, but `GDPRConsentDialog` suffers as a consumer.

Because this component behaves as a modal I think there are a couple things we need to change to make it easier to implement.

## Bad document flow
The root node of `GDPRConsentDialog` is `div` with `position: static` (i.e. the default), even though all the children are removed from the normal document flow (i.e. `position: fixed`) the root node inherits the height and takes up space in the document where you render it.

Demo:
Kapture 2023-08-22 at 16 52 35

## Inflexible container configuration
The default container interacts poorly when implemented in a complex layout, without changes this feels like a deal breaker.

There's a couple of layers to this.

1. Instead of defaulting to a node rendered within the current context, the container should default to using a `<Portal>` with an auto-generated node in the document root.
    - This is a pretty sensible default that popovers usually use.
    - This means things like the "overlay" (the grey bg) and the dialog itself appear above all other content and outside of the normal document flow without any intervention from the consumer.
2. The portal container should be configurable, see mui `<Modal >` [for an example of how this should work](https://mui.com/material-ui/api/modal/#Modal-prop-container). This is really important for supporting complex document layouts.

Here's an example of what happens with the current container settings

Bad rendering result of the overlay (see top right z-fighting + invisible dialog)
image

If I change the overlay z-index to give it priority, the sidebar gets covered which is undesireable
image

What I actually want (I configured the container to respect the layout, and appear on the left)
image

ahmed-gz:
> This feedback is mostly for the underlying `Dialog`, but `GDPRConsentDialog` suffers as a consumer.
> 
> Because this component behaves as a modal I think there are a couple things we need to change to make it easier to implement.
> 
> ## Bad document flow
> The root node of `GDPRConsentDialog` is `div` with `position: static` (i.e. the default), even though all the children are removed from the normal document flow (i.e. `position: fixed`) the root node inherits the height and takes up space in the document where you render it.
> 
> Demo: Kapture 2023-08-22 at 16 52 35
> 
> ## Inflexible container configuration
> The default container interacts poorly when implemented in a complex layout.
> 
> There's a couple of layers to this.
> 
> 1. Instead of defaulting to a node rendered within the current context, the container should default to using a `<Portal>` with an auto-generated node in the document root.
>    
>    * This is a pretty sensible default that popovers usually use.
>    * This means things like the "overlay" (the grey bg) and the dialog itself appear above all other content and outside of the normal document flow without any intervention from the consumer.
> 2. The portal container should be configurable, see mui `<Modal >` [for an example of how this should work](https://mui.com/material-ui/api/modal/#Modal-prop-container). This is really important for supporting complex document layouts.
> 
> Here's an example of what happens with the current container settings
> 
> Bad rendering result of the overlay (see top right z-fighting + invisible dialog) image
> 
> If I change the overlay z-index to give it priority, the sidebar gets covered which is undesireable image
> 
> What I actually want (I configured the container to be in a place where it interacts well with the content area on left) image
> 
> Any thoughts @ahmed-gz @ikovic?

@jake-slashid thanks a lot for the detailed feedback, I hadn't tested the dialog in a complex layout honestly so I didn't expect it to break like that. It should be possible to expose a prop for a custom portal container which is already supported by [Radix UI](https://www.radix-ui.com/primitives/docs/components/dialog#custom-portal-container) but I'm not sure how we can handle the theme correctly because we need the portal container to be a descendent of the `<ThemeRoot>` component for the CSS variables to be pickup up.
The only reason I attached the portal container to a `div` parent of the `<Dialog>` component is to make sure the theme is inherited correctly, any ideas on how to handle this? we could expose a portal container prop but it needs to be wrapped with a `<ThemeRoot>` component, what do you think @ikovic @jake-slashid ?

jake-slashid:
I think the `defaultOpen` prop is confusing 🤔 

I was surprised to see the dialog opened anyway when I have it set to `false`, I have no consent set currently. I think the implicit behaviour is mostly why its confusing.

It might be better to have a "mode" type prop instead, so the behaviour can be chosen:
```tsx
<GDPRConsentDialog open="always" />
// equivalent to defaultOpen=true

<GDPRConsentDialog open="first-time" />
// the current implicit behaviour

<GDPRConsentDialog open={undefined} />
// it's closed, user must click to open (default)
```

The last example is only for demonstration purposes, the actual default would be:
```tsx
<GDPRConsentDialog />

// i.e. without any setting, it defaults to closed
```

jake-slashid:
@ahmed-gz @ikovic might be good for us to have a call tomorrow to talk through some of these things, I've probably posted too much here to keep the conversation productive 😅

jake-slashid:
After pressing "customize" it's a UX dead-end, I can't go back

image

ahmed-gz:
> 

@jake-slashid Sorry I just realized how confusing the logic of the default open state is when `defaultOpen` prop used while there is no consent 😅 , I agree that the prop should allow the user to have more control over the open state in a predictable way rather than overriding it with internal logic.

ahmed-gz:
> Sorry if this ends up being a bit spammy - I'll be posting my thoughts here as I progress through an implementation of the component in my demo app
> 
> 1. `GDPRConsentDialog` needs to be exported in `main.ts`

@jake-slashid @ikovic Should I export the `useGdprConsent` hook as well or should it be for internal use only?

jake-slashid:
Question I had: if the component is in modal mode, does it make sense for it to be dismissable? When I click outside the dialog it closes.

My thoughts:
1. This is probably bad UX because as a brand new user whos never seen it before it's very easy to close it by accident without setting your cookie settings
2. For use cases where the consumer would actually _want_ it to be a modal, I expect they'd be doing so because they probably want to block the user from progressing until they have made a choice?

ikovic:
> > After pressing "customize" it's a UX dead-end, I can't go back
> > image
> 
> @jake-slashid Nice catch, we didn't have the use case to go back to the previous state in the figma design, currently you can go back to the initial state by closing then opening the dialog. @ikovic I guess we will need to ask Ines to add the design of this use case right?

IMO we should add the reject all in the expanded state, then there is no reason to have a button to switch back to the initial state. Ines is not available this week, we can ask for her thoughts on this next week. cc @jake-slashid

ikovic:
> > Sorry if this ends up being a bit spammy - I'll be posting my thoughts here as I progress through an implementation of the component in my demo app
> > 
> > 1. `GDPRConsentDialog` needs to be exported in `main.ts`
> 
> @jake-slashid @ikovic Should I export the `useGdprConsent` hook as well or should it be for internal use only?

If the hook gives you complete control over setting the cookies, I would expose it. Eg the dialog component should only have the UI, the whole logic related to storing and fetching consent should be in the hook.

jake-slashid:
Outcome of portal container discussion:

1. We need to move the "default theme root" (i.e. the css variables) to the `body` element
    - We will continue to allow use of `ThemeRoot` which will implement a scoped theme - this allows sub theme (light -> dark -> light, etc)
2. The `GDPRConsentDialog` will by default mount in a portal at the root
3. Optionally the "container" can be configured via a `container` prop similar to mui `<Modal>`


re: #2
```html
<body>
  <main>
    <SlashIDProvider>
       <!-- your app code -->
    </SlashIDProvider
  </main>
  <div id="some-id">
    <!-- dialog/modal appears here -->
  </div>
</body>
```

jake-slashid:
@ahmed-gz I think your hunch was right, we did miss a use case with `forceOpen` + `forceConsent`!

Based on the change we discussed I think this implementation will show nothing:
```tsx
<GDPRConsentDialog />
```

You might have already thought of this, but just in case: we need to consider the case where no props are provided but there is no consent.

```ts
const isOpen = forceOpen || noConsent
```

ahmed-gz:
> reject all

@jake-slashid @ikovic I like the idea of showing reject all in the expanded state as well, the only question I have is should it have the same variant as the accept all button or save settings button? I believe the data protection officer feedback included something about the reject all not being highlighted less than the other buttons

ikovic:
> > reject all
> 
> @jake-slashid @ikovic I like the idea of showing reject all in the expanded state as well, the only question I have is should it have the same variant as the accept all button or save settings button? I believe the data protection officer feedback included something about the reject all not being highlighted less than the other buttons

Yes, accept all and reject all should always have the same "importance"

ahmed-gz:
> main

@jake-slashid To make it easier, here is a table with all possible combinations, let me know if there is an inconsistency:
| `forceOpen` | `forceConsent` | `noConsent` | **Resulting Dialog Open State** | **Comments/Questions** |
|--------|--------|--------|--------|--------|
| `true` | **N/A** | **N/A** | `true` | **N/A** here just means those values don't matter in this case |
| `false` | `true` | `true` | `true` | - |
| `false` | `true` | `false` | `false` | In this case I think we should hide dialog even with `forceConsent=true` because user already has stored consents |
| `false` | `false`  | `true` | `true` | here we just default to `noConsent` |
| `false` | `false`  | `false` | `false` | here we just default to `noConsent` |

ahmed-gz:
> dismiss

@jake-slashid yes `noConsent` in the table means `noConsent = !consents.length`.
Ok so it looks like we have everything covered related to this, thanks a lot!

packages/react/src/components/gdpr-consent-dialog/actions.tsx
```diff
@@ -0,0 +1,122 @@
...
+        label="Accept all"
+        hasError={hasError}
+        isActive={activeAction === "accept"}
+        loading={isLoading}
+        onClick={handleAccept}
+        className={styles.acceptButton}
+      />
+      <ActionButton
+        testId="sid-gdpr-consent-dialog-reject"
+        variant={showSaveButton ? "neutralMd" : "secondaryMd"}
```
jake-slashid, line 102:
```suggestion
        variant="secondaryMd"
```

I think there was a small misunderstanding here

Ivan said accept all + reject all should always be the same

image

packages/react/src/components/gdpr-consent-dialog/types.ts
```diff
@@ -14,22 +14,20 @@ export type GDPRConsentDialogProps = {
...
-  defaultOpen?: boolean;
   /** Value of the lock on the necessary cookies category */
   necessaryCookiesRequired?: boolean;
   /** Default consent levels to store when clicking on accept all */
   defaultAcceptAllLevels?: ConsentSettingsLevel[];
   /** Default consent levels to store when clicking on reject all */
   defaultRejectAllLevels?: GDPRConsentLevel[];
+  /** When this boolean is set to `true`, it disables interactions outside the dialog and prevents the user from closing it until they choose an option */
+  forceConsent?: boolean;
+  /** When this boolean is set to `true`, it forces the dialog to be open on page load */
```
jake-slashid, line 25:
```suggestion
  /** When this boolean is set to `true`, it forces the dialog to be open on page load regardless of if consent was given previously */
```

jake-slashid (approved):
From a scan of the code I think it looks great, nice job!

I'll try it in my test environment again today but should be good 👍

packages/react/src/components/gdpr-consent-dialog/actions.tsx
ahmed-gz: thanks for catching that, I don't know why I read accept as save 😅

packages/react/src/components/theme-root/index.tsx
```diff
@@ -19,25 +20,26 @@ type Props = {
...
-          [autoTheme]: theme === "auto",
-          [lightThemeVars]: theme === "light",
-        },
-        className
-      )}
-    >
-      {children}
-    </div>
-  );
+  useLayoutEffect(() => {
```
ikovic, line 28: @ahmed-gz @jake-slashid what do you think? Tested it locally and it seems to behave as expected.

PR #119, opened 2023-08-04
[SID-1137] add more tests for OrganizationSwitcher

jake-slashid:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1137)

## Changes
- Add tests for `OrganizationSwitcher`
- Refactor `() => Promise.resolve(...)` to `async () => ...`
- Fix broken test for `MultiFactorAuth`
- Some chores to satisfy the linter

## Checklist


packages/react/src/components/organization-switcher/organization-switcher.test.tsx
```diff
@@ -62,49 +58,196 @@ describe.skip("OrganizationSwitcher", () => {
...
+    
+    await waitFor(() => expect(screen.queryByText("-")).not.toBeInTheDocument())
+    await waitFor(() => expect(container.querySelector(".sid-organization-switcher > .sid-dropdown .sid-dropdown__trigger")).toBeTruthy())
+
+    const trigger = container.querySelector(".sid-organization-switcher > .sid-dropdown .sid-dropdown__trigger")!!
+    
+    ;(window as any).PointerEvent = class PointerEvent extends Event {};
+    window.HTMLElement.prototype.scrollIntoView = vi.fn();
+    window.HTMLElement.prototype.hasPointerCapture = vi.fn();
+    window.HTMLElement.prototype.releasePointerCapture = vi.fn();
```
jake-slashid, line 146: This is nasty but I don't know what else to do about it:
https://github.com/testing-library/user-event/discussions/1087

ikovic: It is what it is - maybe leave a comment with the source?

ikovic (approved):

jake-slashid:
I'm moving on but want to make note of this:

With our current test setup the middleware prop of `<Form />` is basically untestable.

The middleware invocation business logic is an implementation detail of `<SlashIDProvider />`, and it's heavily coupled with `@slashid/slashid` which as far as I can tell is not easy to mock or stub in any terse way. Using the `<TestSlashIDProvider />` is not useful here.

Without being able to stub `@slashid/slashid` we can't test the following cases:
```ts
test("should call each middleware with the outcome of the previous" ...)
test("should provide the initial user to the middleware layer" ...)
test("should set the middleware-ified user in SlashIDProvider" ...)
```

jake-slashid:
That said I guess we could abstract the middleware business logic itself so it's testable, I would still like some guarantee though that when I release the SDK that the `middleware` prop behaves as it should - it seems quite important.

PR #126, opened 2023-08-11
[SID-1173] Allow string as arg for Groups.belongsTo

jake-slashid:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1173)

- `belongsTo` accepts `string | (groups: string[]) => boolean`
- Rewrite the tsdoc string
- Documentation: https://github.com/slashid/ng-evangelion/pull/958

## Checklist


packages/react/src/components/groups/groups.test.tsx
```diff
@@ -43,6 +44,48 @@ describe("Groups", () => {
...
 
+  test("should render children when the SDK is ready and the user belongs to the specified groups, as string", () => {
+    const userWithGroups = createTestUser()
+    const groups = Array
+      .from(Array(faker.number.int({ min: 3, max: 10 })))
+      .map(() => faker.company.buzzNoun())
+
+    userWithGroups.getGroups = vi.fn(() => groups);
+
+    const [belongsTo] = faker.helpers.shuffle(groups)
```
ikovic, line 55: What is the added value of using faker here vs a predefined set of constants? Personally I find this harder to read so the live documentation aspect of tests is a bit hurt. Is it a good trade off?

packages/react/src/components/groups/index.tsx
```diff
@@ -4,13 +4,50 @@ import { useSlashID } from "../../main";
 type Groups = string[];
 
 type Props = {
-  belongsTo: (groups: Groups) => boolean;
+  belongsTo: string | ((groups: Groups) => boolean);
```
ikovic, line 7: Does it make sense to support `string[]` while we're at it?

ikovic: Just saw the comment you made in the ticket description - I think it wouldn't be surprising to treat the `string[]` option as `AND`. Any more complex logic could be handled using the callback. What do you think?

packages/react/src/components/groups/groups.test.tsx
jake-slashid: When we use static data we risk introducing false positives in tests.

Here's an example scenario:
1. The group `groupsa` is used to write a test when the `<Groups>` component is created, it's carefully double checked to assure it fails when the value is changed.
2. The test gives us peace of mind when making changes because it should guarantee this particular feature works. It's tested.
3. We change `<Groups>` some time later, accidentally we introduce an edge case where if exactly `groupsa` is provided as the prop, this test _always_ passes but otherwise the feature is actually broken. We've introduced a magic string basically.
4. We assume the component works because the test has always passed and is still passing, it must be healthy. Publish.
5. 💥

By introducing some randomness we prevent these kind of magic values and increase the robustness of the test, without it we're testing that specific values work as opposed to testing that the data flows correctly.

Trade off: we lose some readability, agree!

jake-slashid:
Documentation: https://github.com/slashid/ng-evangelion/pull/958

packages/react/src/components/groups/index.tsx
jake-slashid: I also have some sample snippets in the docs which elaborate on the XOR/AND thing a bit more:
https://github.com/slashid/ng-evangelion/blob/e8314e059c1f851e822680dbfcdd5e6087786538/web/dev-portal/docs/react-sdk/reference/components/groups.md

jake-slashid: Oh I only just seen your second comment 😓 Github didn't update or something

I think AND being default for an array is probably okay, do you think AND is more common than OR?

paulo-raca: I`m a bit confused by XOR... Do you mean regular OR?

Anyway, I think it makes sense to support multiple groups easily. Suggestions:
- Modify this to be `callback | string | (string | string[])[]`. The first array level maps to an OR and the second maps to AND. A few examples:
  - callback works as usual
  - `"group1"`, `"[group1]"` and `"[[group1]]"` are sinonimous and checks if the user belongs to group1
  - `["group1", "group2"]` and`[["group1"], ["group2"]]` are sinonimous and checks if the user belongs to either group1 or group2
  - `[["group1", "group2"]]` checks if the user belongs to both group1 and group2
  - `["group1", ["group2", "group3"]]` checks if the user belongs group1 or (it belongs to both group2 and group3)
  - Stretch goal: while at it, maybe support "!group", so that this becomes a complete [DNF](https://en.wikipedia.org/wiki/Disjunctive_normal_form)-like syntax :joy: 
- Instead, maybe add `belongsToAny` and `belongsToAll` props? 
  Not as flexible, but very self-explanatory.

paulo-raca: I just noticed this suggestion:

```
<Groups
  belongsTo={all(a, b, c)}
  belongsTo={some(a, b, c)}
>
```

I like this a lot! Extra points if you can mix them: `belongsTo={any(all("a", "b"), "c", not("d")}`

ikovic: @paulo-raca thanks for the comment! I would lean towards the most explicit interface, reducing the implicit behaviour to minimum. I honestly do now know what the primary/most common use cases would be though 🤔

packages/react/src/components/groups/groups.test.tsx
ikovic: Hmm I think we could have both, right? Do you think it would be too redundant to add a more streamlined test with a constant just for the live docs purpose? With that we solved both problems 😃

packages/react/src/components/groups/index.tsx
jake-slashid: Yes I meant OR 🤦

jake-slashid: If we have to choose a default behaviour I prefer the explicitly named inflexible props, they're much more self documenting and simple to grok

```
<Groups
  belongsTo="a"
  belongsTo={groups => groups.includes("a")}
  belongsToAny={[a, b]}
  belongsToAll={[a, b])
/>
```

We can throw if more than one is set

packages/react/src/components/groups/groups.test.tsx
ikovic: It would essentially be the same, just without the iteration and generating group names. So a shorter, more readable version. I don't think it's too important though given that it's not a complex component/test, please proceed in any way you think makes sense!

jake-slashid: Lets do both, I think it makes sense

jake-slashid:
This is quite low in my priorities, I should come back to it by end of this week

packages/react/src/components/groups/example.tsx
```diff
@@ -0,0 +1,47 @@
...
+      belongsTo={"hello"}
+
+      belongsToAny={["hello", "world"]}
+
+      belongsToAll={["hello", "world"]}
+    >
+      hello world
+    </Groups>
+  )
+}
```
jake-slashid, line 47: Please see these prototypes!

ikovic: I can't rely on anything but personal preference here - I much prefer the Util example:
- smaller API surface
- no need to assert only a single mutually exclusive prop was used
- no implicit behaviour
- no room for typos (`belongsToAny` and `belongsToAll` might be swapped mistakenly)

I also don't know what the most frequent use cases are going to be, but I do have a feeling that the most commonly used option would be just specifying a single group name.

jake-slashid: I would prefer to have them exported like
```
import { Groups } from '@slashid/react'
import { any, all } from '@slashid/react/groups'

// or

import { Groups, any, all } from '@slashid/react'
```

but this feels like a pretty good compromise

jake-slashid:
added tests, removed new props

ikovic (approved):

Issue #125, opened 2023-08-11
Add storybook


ahmed-gz:
Add storybook to the `react` package and document components exported from `main.ts`

PR #106, opened 2023-07-27
[SID-1158] Standardise children prop type

jake-slashid:
Dependency: https://github.com/slashid/javascript/pull/105

## Description
[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1158)

I had to make a decision about which type to use when building `<OrganizationSwitcher />`, this PR brings consensus to the SDK.

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have generated the new version of the docs website and smoke tested it
- [ ] I have checked that my changes haven't caused semantic errors in the existing docs
- [ ] I have updated the README and DEVELOPMENT files

PR #127, opened 2023-08-15
[Issue-125] Setup storybook for react package

ahmed-gz:
## Description

[Github issue](https://github.com/slashid/javascript/issues/125)

Added storybook in the react package along with stories for the components we expose to customers which are exported from `main.ts`.

To run `storybook` first `cd packages/react` then run `pnpm storybook`.

## Checklist


packages/react/src/components/form/form.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+  component: Form,
+};
+
+export default meta;
+
+type Story = StoryObj<typeof Form>;
+
+export const Primary: Story = {
+  render: () => (
+    <ConfigurationProvider>
```
jake-slashid, line 15: I don't think that the configuration provider is mandatory, so this feels extraneous/confusing for a new user.

Maybe we should remove it?

packages/react/src/components/form/form.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+      <Form
+        middleware={({ user }) => {
+          console.log("middleware - user: ", user);
+          return new Promise((resolve) => {
+            resolve(user);
+          });
+        }}
```
jake-slashid, line 22: I think we should probably remove this, it's quite an advanced setting and doesn't feel like it's contributing much here.

packages/react/src/components/form/form.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+        factors={[{ method: "email_link" }, { method: "otp_via_sms" }]}
+        text={{
+          "initial.title": "Multi-Factor Authentication",
```
jake-slashid, line 25: Similar to other feedback, as the default story this feels out of place.

I would suggest that for the default story we show the absolute minimum implementation which is:
```ts
<Form />
```

What do you think?

packages/react/.storybook/preview.tsx
```diff
@@ -0,0 +1,26 @@
...
+    },
+  },
+};
+
+export const decorators = [
+  (Story) => (
+    <SlashIDProvider
+      // @ts-ignore
+      oid={import.meta.env.VITE_ORG_ID}
+      themeProps={{ theme: "dark" }}
```
jake-slashid, line 19: I think here we should use the default (`auto`) but expose this setting via the [controls plugin](https://storybook.js.org/docs/react/essentials/controls) somehow.

It would be useful for people to know and see it can be changed easily, but should default to their OS/browser setting.

If controls plugin is too complicated, I would still suggest that we use `auto` at the very least.

jake-slashid (commented):
I think it's looking great!

I have two things to ask:
1. Could you check the [latest version of the reference documentation](https://developer.slashid.dev/docs/react-sdk) and confirm that the examples here match or are close to the examples there, and they we have them all? For example `LoggedOut` has an example in the docs which is not here.
2. As a general overarching piece of feedback: I would suggest that for every component the primary example is the bare minimum implementation, then for each "variant" (i.e. unique combination of props) we have another example.

The sidebar might look something like this for `\<Form>`:
```
- Default
- With success callback
- With custom factors
- With text override
- With middleware
```

Then at a glance it's possible to see everything the component has to offer.

packages/react/src/components/logged-in/logged-in.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+export const Primary: Story = {
+  render: () => (
+    <LoggedIn withFactorMethods={["email_link", "otp_via_sms"]}>
+      Only visible for the users authenticated with email magic link and SMS
+      code.
+    </LoggedIn>
+  ),
+};
+
+export const Secondary: Story = {
```
jake-slashid, line 21: I think that primary/secondary made sense for the button example because it was showing the primary and secondary buttons, but here it feels like without opening these stories I wouldn't know what they contain.

I'm not sure how to define custom story names with this object pattern but when defining a story using function you can do it like this:
```tsx
const Secondary = () => <LoggedIn withFactoryMethods={...} />
Secondary.storyName = "With factor count check"
```

Do you think we can improve discoverability of stories like this?

packages/react/src/components/form/form.stories.tsx
kaspermroz: Agree with Jake here, initial stories should have bare minimum, ideally only required props. We can add more stories to showcase advanced config options later

packages/react/src/components/form/form.stories.tsx
ahmed-gz: yes you're right, I will change it 👍

packages/react/src/components/logged-in/logged-in.stories.tsx
ahmed-gz: yes I agree, thanks for the suggestion 👍

packages/react/src/components/form/form.stories.tsx
ahmed-gz: Yes that makes more, thank you!

ahmed-gz: @jake-slashid I just checked the [docs](https://developer.slashid.dev/docs/react-sdk/reference/components/react-sdk-reference-form) and it say that **_a `<ConfigurationProvider>` needs to be rendered as a parent component_**, doesn't that mean that it's mandatory for the `Default` story?

jake-slashid: I'm building a demo at the moment and thought I did it without the configuration provider, looking at the source code its pretty clear the configuration context is required.

We should probably change this but for now, please ignore me and include it in the example!

ahmed-gz: @jake-slashid I tried using the same example for `<SlashIDLoaded>` [here](https://developer.slashid.dev/docs/react-sdk/reference/components/react-sdk-reference-slashidloaded) but it says that it doesn't accept text as child elements:
```
    <SlashIDLoaded>
      SlashID is ready!
    </SlashIDLoaded>
```

Should we update the docs or change the children type to `React.ReactNode`?

jake-slashid: There's a PR for it here: https://github.com/slashid/javascript/pull/106

packages/react/.storybook/preview.tsx
ahmed-gz: I tried to expose the theme setting but it seems that it's only possible for story component prop, so I'm not sure how to do it for `<SlashIDProvider>` since it's used as a wrapper for every story, any idea how to make it work?

ikovic (approved):

PR #129, opened 2023-08-18
[SID-1186] Reset the Org ID and SID instance on logout

kaspermroz:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1186)

### The bug
We have a bug in the React SDK. Steps to reproduce:
 - start anew, new tab, clean storage etc
 - register with a new address
 - create a suborg and switch to it
 - logout without closing the tab
 - you'll be presented with the form again, register with another new address

=> **the request to /id goes out with

 - the token for the previously logged-in user
 - the OrgID of the suborg, not the demo root**

### Solution
After logging out, create a new SlashID instance with the root organization ID and no token.

## Checklist

### DIFF
--- a/packages/react/src/context/slash-id-context.tsx
+++ b/packages/react/src/context/slash-id-context.tsx
@@ -10,7 +10,7 @@ import React, {
 import { PersonHandleType, SlashID, User } from "@slashid/slashid";
 import { MemoryStorage } from "../browser/memory-storage";
 import { LogIn, MFA } from "../domain/types";
-import { SDKState, sdkStates } from "../domain/sdk-state";
+import { SDKState } from "../domain/sdk-state";
 import { ThemeProps, ThemeRoot } from "../components/theme-root";
 export type StorageOption = "memory" | "localStorage";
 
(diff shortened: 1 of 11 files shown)


jake-slashid (approved):

jake-slashid (approved):

jake-slashid:
Will the tests continue to pass here after https://github.com/slashid/ng-evangelion/pull/981 is merged?

As part of upgrading to that version I'll remove these lines:
https://github.com/slashid/javascript/pull/129/files#diff-6012c9cb5e484ecbab997929767ddbcbdaaac00ea8087b7ace4bf47424f740ceR126-R136

packages/react/src/context/slash-id-context.tsx
```diff
@@ -123,7 +123,17 @@ export const SlashIDProvider: React.FC<SlashIDProviderProps> = ({
 
     user.logout();
     setUser(undefined);
-  }, [user, state]);
+    // we need to set the oid back to the root on log out
+    setOid(initialOid);
+    // we can't reinitialize the SDK here, as it picks up the old token and creates a new user immediately
+    sidRef.current = new SlashID({
```
ikovic, line 129: @kaspermroz would it make sense to wrap this behind a factory function like `createSlashID(options)`? Looks like it happens in two places and there is some logic behind the options and default values.

PR #132, opened 2023-08-23
[SID-1199] Add the otp_via_email auth method

ikovic:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1199)

Follow up work to get the `otp_via_email` method working.

## Checklist


packages/react/src/domain/handles.ts
```diff
@@ -1,12 +1,12 @@
...
 import { Factor } from "@slashid/slashid";
 import {
   getList,
   findFlagByDialCode,
 } from "country-list-with-dial-code-and-flag";
 import { TextConfigKey } from "../components/text/constants";
 
-const FACTORS_WITH_EMAIL = ["webauthn", "email_link"];
+const FACTORS_WITH_EMAIL = ["webauthn", "otp_via_email", "email_link"];
 const FACTORS_WITH_PHONE = ["otp_via_sms", "sms_link"];
```
paulo-raca, line 10: :thinking: IIRC we can do WebAuthn + SMS

jake-slashid (approved):

packages/react/src/dev.tsx
```diff
@@ -157,8 +173,14 @@ ReactDOM.createRoot(document.getElementById("root") as HTMLElement).render(
     >
       <div className="layout">
         <div>
-          <h2>Switch to default org</h2>
-          <Config />
+          <div>
```
paulo-raca, line 178: Is this a test page?

paulo-raca (approved):
LGTM, thank you

balenio (approved):

packages/react/src/dev.tsx
ikovic: Correct, this is our development page where we render all the different app configs. Not exposed to customers in any way 😄

packages/react/src/domain/handles.ts
ikovic: I think we can but we don't support it for some reason I cannot remember right now 😓

PR #133, opened 2023-08-23
[SID-1202] Use correct OIDC branding

jake-slashid:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1202)

Solving the same problem as https://github.com/slashid/ng-evangelion/pull/982

## Checklist

### DIFF
--- a/packages/react/src/components/form/initial/oidc.tsx
+++ b/packages/react/src/components/form/initial/oidc.tsx
@@ -1,4 +1,4 @@
-import { Factor } from "@slashid/slashid";
+import { Factor, OAuthProvider } from "@slashid/slashid";
 import clsx from "clsx";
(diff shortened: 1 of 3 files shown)


packages/react/src/components/form/initial/oidc.tsx
```diff
@@ -15,9 +15,9 @@ import { Line } from "../../icon/line";
 
 import * as styles from "./initial.css";
 
-const PROVIDER_TO_ICON: Record<string, React.ReactNode> = {
+const PROVIDER_TO_ICON: Record<OAuthProvider, React.ReactNode> = {
```
jake-slashid, line 18: This type change means we get build errors if a new provider is added which we don't have a pretty name or icon for.

ikovic (approved):

PR #135, opened 2023-08-23
[SID-1190] Upgrade to @slashid/slashid@3.13.2

jake-slashid:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1190)

Fixes bug with org switching and stale SDK state after logout

## Checklist

### DIFF
--- a/packages/react/src/context/slash-id-context.tsx
+++ b/packages/react/src/context/slash-id-context.tsx
@@ -125,14 +125,7 @@ export const SlashIDProvider: React.FC<SlashIDProviderProps> = ({
     setUser(undefined);
     // we need to set the oid back to the root on log out
     setOid(initialOid);
-    // we can't reinitialize the SDK here, as it picks up the old token and creates a new user immediately
-    sidRef.current = new SlashID({
-      oid: initialOid,
-      ...(baseApiUrl && { baseURL: baseApiUrl }),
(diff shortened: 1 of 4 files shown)


ikovic (approved):

packages/react/src/context/slash-id-context.tsx
```diff
@@ -125,14 +125,7 @@ export const SlashIDProvider: React.FC<SlashIDProviderProps> = ({
     setUser(undefined);
     // we need to set the oid back to the root on log out
     setOid(initialOid);
-    // we can't reinitialize the SDK here, as it picks up the old token and creates a new user immediately
-    sidRef.current = new SlashID({
```
ikovic, line 129: The SDK instance is created with an `oid` - when we log out, we won't reset this OID, but when we switch orgs we create a new SDK instance with the new OID. Am I missing something, do we have another way of clearing this value now?

ikovic: @jake-slashid @kaspermroz looking in more detail, passing `oid` to the SlashID constructor is inconsequential - that value is not being used. I remember adding it there as a stepping stone to the next major version. We still pass the oid as a mandatory parameter to `sdk.id()` every time we log in.

kaspermroz (approved):

jake-slashid:
Bumping to include incoming version `3.13.2`

PR #136, opened 2023-08-24
[SID-1206] Add renderLabel prop to <OrganizationSwitcher>

jake-slashid:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1206)

Two changes
- Add `renderLabel` prop to `<OrganizationSwitcher>`, see context.
    - I'm changing `Dropdown` to allow `ReactNode` for `label`
    - The type is now `{ value: string, label: ReactNode }`
- Add npm script `next`. It automatically publishes things to `yalc` with a bumped pre-release version.

### Context
The org list currently renders org names "as is" which can create some interesting problems for consumers when weighing up the options for choosing a programmatically generated organization name scheme.

This change proposes allowing org name overrides (i.e. display names) in the most prominent place the org name is shown: the org switcher.

This feature gives flexibility to consumers to choose internal-only names for programmatically generated organizations, or even to implement internationalisation on org names if they wish.

```tsx
<OrganizationSwitcher
  renderLabel={org => {
    if (org.id === "some_id") return "Some custom org name!"
    return org.org_name
  }}
/>
```

```tsx
<OrganizationSwitcher
  renderLabel={org => {
    return i18n(org.id)
  }}
/>
```

## Checklist

### DIFF
--- a/.changeset/cyan-mangos-judge.md
+++ b/.changeset/cyan-mangos-judge.md
@@ -1,0 +1,5 @@
+---
+"@slashid/react": minor
+---
+
+Add renderLabel prop to \<OrganizationSwitcher />
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -43,7 +43,8 @@
     "lint": "eslint ./src",
     "test:types": "tsc --noEmit --emitDeclarationOnly false",
     "test:unit": "vitest run",
-    "test:unit:watch": "vitest"
+    "test:unit:watch": "vitest",
+    "next": "node ./yalc.next.js && pnpm build && yalc publish"
   },
   "dependencies": {
     "@radix-ui/react-dialog": "^1.0.4",
@@ -83,10 +84,12 @@
     "msw": "^1.2.3",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
+    "semver": "^7.5.4",
     "storybook": "7.0.0-rc.2",
     "tsconfig": "workspace:*",
     "vite": "^3.1.0",
-    "vitest": "^0.25.2"
+    "vitest": "^0.25.2",
+    "yalc": "1.0.0-pre.53"
   },
   "peerDependencies": {
     "@slashid/slashid": ">= 3.13.0",
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -44,7 +44,7 @@
     "test:types": "tsc --noEmit --emitDeclarationOnly false",
     "test:unit": "vitest run",
     "test:unit:watch": "vitest",
-    "next": "node ./yalc.next.js && pnpm build && yalc publish"
+    "publish:yalc": "node ./yalc.next.js && pnpm build && yalc publish"
   },
   "dependencies": {
     "@radix-ui/react-dialog": "^1.0.4",
--- a/packages/react/src/components/organization-switcher/index.tsx
+++ b/packages/react/src/components/organization-switcher/index.tsx
@@ -10,6 +10,7 @@ import { OrganizationDetails } from "@slashid/slashid";
 interface Props {
   fallback?: ReactNode;
   filter?: (organization: OrganizationDetails) => boolean;
+  renderLabel?: (organization: OrganizationDetails) => ReactNode;
 }
 
 const className = sprinkles({ marginBottom: "3", marginTop: "5" });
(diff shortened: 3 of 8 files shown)


packages/react/src/components/organization-switcher/index.tsx
```diff
@@ -10,6 +10,7 @@ import { OrganizationDetails } from "@slashid/slashid";
 interface Props {
   fallback?: ReactNode;
   filter?: (organization: OrganizationDetails) => boolean;
+  renderLabel?: (organization: OrganizationDetails) => ReactNode;
```
jake-slashid, line 13: I intentionally did not name this `renderItem` because it does not allow you to control the full rendering of the menu item only the label.

The selected item for example has a checkmark, we don't allow you to configure this (yet).

I anticipate in future we would also have `renderItem` which would allow control of the entire item renderer, but out of scope for now.

jake-slashid (commented):
Here is a practical example of how this feature is useful:

In this multi-tenancy demo I can present "Your personal organization" to every user, greatly simplifying the UX

image

packages/react/package.json
```diff
@@ -43,7 +43,8 @@
     "lint": "eslint ./src",
     "test:types": "tsc --noEmit --emitDeclarationOnly false",
     "test:unit": "vitest run",
-    "test:unit:watch": "vitest"
+    "test:unit:watch": "vitest",
+    "next": "node ./yalc.next.js && pnpm build && yalc publish"
```
kaspermroz, line 47: I'd think of a slightly more meaningful name for it, `next` makes me think of Next.js. How about `next-version` or `prerelease` ?

kaspermroz: or even `yalc` / `yalc-publish`

ikovic: Following the conventions we had so far, my vote would be `publish:local` or `publish:yalc`

ikovic:
@jake-slashid the deployment on cloudflare complains about a broken lockfile but it looks good to me 🤔  it might be using another version of PNPM though

jake-slashid:
@ikovic the lockfile was broken when I first pushed, it's green now.

Does it auto deploy on push or can we trigger it again somehow?

ikovic:
> @ikovic the lockfile was broken when I first pushed, it's green now.
> 
> Does it auto deploy on push or can we trigger it again somehow?

Last build was executed on top of [a96d101](https://github.com/slashid/javascript/pull/136/commits/a96d10126d24f7cfe95497fe266f619506497f0a). I'll retry the deployment from the Cloudflare dashboard

jake-slashid:
Docs: https://github.com/slashid/ng-evangelion/pull/995

packages/react/package.json
jake-slashid: I renamed it `publish:yalc`

jake-slashid:
Ready for final review!

- Added storybook example
- Added tests
- Wrote docs: https://github.com/slashid/ng-evangelion/pull/995
- Addressed PR feedback
- CI green ✅

ikovic (approved):

kaspermroz (approved):
Just couple minor comments, other than that it looks good! 🎉

.changeset/cyan-mangos-judge.md
```diff
@@ -0,0 +1,5 @@
+---
+"@slashid/react": minor
+---
+
+Add renderLabel prop to \<OrganizationSwitcher />
```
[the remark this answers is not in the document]
jake-slashid: It has to be here because github renders HTML 

Here's an example: \<Hello />
Here's an example: <Hello />

packages/react/src/components/organization-switcher/organization-switcher.test.tsx
```diff
@@ -139,6 +139,38 @@ describe("OrganizationSwitcher", () => {
...
+          renderLabel={(org) => {
+            if (org.id === currentOrg.id) return expected;
+            return org.org_name;
+          }}
+        />
+      </TestSlashIDProvider>
+    );
+
+    await waitFor(() =>
+      expect(screen.queryByText("-")).not.toBeInTheDocument()
```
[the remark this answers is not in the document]
jake-slashid: You're right but I couldn't think of another way to do it, because the loading state is actually the same component as the rendered state - there are no differences.

I think testing for the current value is probably a better idea, let me try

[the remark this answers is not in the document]
jake-slashid: Done! It's a bit noisy but more future proof

PR #141, opened 2023-08-28
[SID-1196] Improve GDPRConsentDialog usability

jake-slashid:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1196)

I encountered some issues trying to implement the component in my test bench.

### Issues
1. Implementing the `container` prop is quite clunky, and the prop doesn't have parity with the `mui/modal` implementation which was the reference
2. Dismiss behaviour not as expected, dismiss by clicking background should only happen in modal mode
3. The advanced mode ("Customize") has truncated buttons because the element isn't wide enough
4. The cookie button does not appear in the portal, so it still suffers the same problem as the dialog did before

### Changes
- `<Dialog />`
    - Improve the `container` prop API, bringing it into parity with `mui/modal`
        - The main improvement here is allowing an (internally cached) getter function for locating the original DOM node, otherwise a lot of complicated boilerplate is delegated to the user.
    -  Component can no longer by dismissed in "dialog mode" by clicking outside the element
        - Dismiss by clicking outside the element is permanently disabled for dialogs i.e. `<Dialog modal={false} />`
        - Dismiss by clicking outside the element is reserved for modals i.e. `<Dialog modal dismissable />` (the default)
        - Dismiss by clicking outside the element can be disabled **for modals only** i.e. `<Dialog modal dismissable={false} />`
- Add `<Teleport />`
    - **Internal use only**
    - This component is a fascade for `React.createPortal()` with a simpler API.
- `<GDPRConsentDialog />`
    - Component is a little wider to allow for all three buttons in the advanced mode
    - The cookie button (shown when the dialog is dismissed) will now render in its own portal, seperate to the dialog content.
        - Without this change the button renders in the tree where the component is mounted, same problem we had before with the dialog itself.
        - I considered putting the button in `<RadixDialog.Portal />` but there is implicit behaviour here where the portal is only shown when `<RadixDialog.Root open={true} />` which is an issue, because this button is only shown when `open=false`.
        - Given all that context, the button is now mounted inside a `<Teleport />` at the document root

#### Examples
Improved `container` prop
```tsx
// now possible

<GDPRConsent
  container={() => document.getElementById("foo")}
/>
```

Teleport
```tsx
<Teleport to="some-name">
  // ...
</Teleport>
```
## Checklist

### DIFF
--- a/packages/react/src/components/dialog/index.tsx
+++ b/packages/react/src/components/dialog/index.tsx
@@ -1,10 +1,10 @@
 import * as RadixDialog from "@radix-ui/react-dialog";
 import { clsx } from "clsx";
-import { ReactNode } from "react";
+import { ReactNode, useMemo } from "react";
 import { Close } from "../icon/close";
 import * as styles from "./style.css";
 
-type Props = {
+export type DialogProps = {
   /** Dialog content */
   children: ReactNode;
   /** Dialog trigger */
@@ -16,7 +16,7 @@ type Props = {
   /** Dialog icon */
   icon?: ReactNode;
   /** Dialog custom portal container */
-  container?: HTMLElement;
+  container?: HTMLElement | (() => HTMLElement | null) | null;
   /** Dialog content container class name */
   className?: string;
   /** When this boolean is set to `false`, the close button is hidden and dialog cannot be closed by clicking on the overlay */
@@ -42,18 +42,33 @@ export const Dialog = ({
   container,
   dismissable = true,
   modal = true,
-}: Props) => {
+}: DialogProps) => {
+  /**
+   * Check if the container function return value changed every cycle but only
+   * trigger another render when it actually changes.
+   */
+  const containerFuncReturn = (() => {
+    if (typeof container !== "function") return null
+
+    return container()
+  })()
+
+  const __container = useMemo(() => {
+    if (typeof container === "function") return containerFuncReturn
+    return container
+  }, [container, containerFuncReturn])
+
   if (!trigger) return null;
 
   return (
     <RadixDialog.Root modal={modal} open={open} onOpenChange={onOpenChange}>
       <RadixDialog.Trigger asChild>{trigger}</RadixDialog.Trigger>
-      <RadixDialog.Portal container={container}>
+      <RadixDialog.Portal container={__container}>
         {modal && <RadixDialog.Overlay className={styles.overlay} />}
         <RadixDialog.Content
           className={clsx("sid-dialog", styles.wrapper, className)}
           onInteractOutside={(e) => {
-            if (!dismissable) {
+            if (!modal || !dismissable) {
               e.preventDefault();
             }
           }}
--- a/packages/react/src/components/dialog/index.tsx
+++ b/packages/react/src/components/dialog/index.tsx
@@ -1,6 +1,6 @@
 import * as RadixDialog from "@radix-ui/react-dialog";
 import { clsx } from "clsx";
-import { ReactNode, useMemo } from "react";
+import { ReactNode } from "react";
 import { Close } from "../icon/close";
 import * as styles from "./style.css";
 
@@ -43,27 +43,12 @@ export const Dialog = ({
   dismissable = true,
   modal = true,
 }: DialogProps) => {
-  /**
-   * Check if the container function return value changed every cycle but only
-   * trigger another render when it actually changes.
-   */
-  const containerFuncReturn = (() => {
-    if (typeof container !== "function") return null
-
-    return container()
-  })()
-
-  const __container = useMemo(() => {
-    if (typeof container === "function") return containerFuncReturn
-    return container
-  }, [container, containerFuncReturn])
-
   if (!trigger) return null;
 
   return (
     <RadixDialog.Root modal={modal} open={open} onOpenChange={onOpenChange}>
       <RadixDialog.Trigger asChild>{trigger}</RadixDialog.Trigger>
-      <RadixDialog.Portal container={__container}>
+      <RadixDialog.Portal container={typeof container === "function" ? container() : container}>
         {modal && <RadixDialog.Overlay className={styles.overlay} />}
         <RadixDialog.Content
           className={clsx("sid-dialog", styles.wrapper, className)}
--- a/packages/react/src/components/gdpr-consent-dialog/consent-dialog.tsx
+++ b/packages/react/src/components/gdpr-consent-dialog/consent-dialog.tsx
@@ -1,10 +1,10 @@
 import { GDPRConsent } from "@slashid/slashid";
 import { clsx } from "clsx";
-import { forwardRef, useEffect, useMemo, useReducer } from "react";
+import { useEffect, useMemo, useReducer } from "react";
 import { publicVariables } from "../../theme/theme.css";
-import { Button } from "../button";
 import { Dialog } from "../dialog";
 import { Cookie } from "../icon/cookie";
+import { TriggerButton } from './trigger-button'
 import { Text } from "../text";
 import { Actions } from "./actions";
 import { CONSENT_LEVELS_WITHOUT_NONE } from "./constants";
@@ -16,7 +16,6 @@ import {
   GDPRConsentDialogProps,
   UpdateGdprConsent,
 } from "./types";
-import { Teleport } from "../teleport";
 
 type Props = GDPRConsentDialogProps & {
   consents: GDPRConsent[];
@@ -69,25 +68,6 @@ export const ConsentDialog = ({
     }
   }, [open]);
 
-  const TriggerButton = forwardRef<HTMLButtonElement>((props, ref) => (
-    <Teleport to="sid-cookie-button">
-      <Button
-        {...props}
-        ref={ref}
-        testId="sid-gdpr-consent-dialog-trigger"
-        variant="neutralMd"
-        className={clsx(
-          "sid-gdpr-consent-dialog-trigger",
-          styles.dialogTrigger,
-          triggerClassName
-        )}
-      >
-        <Cookie />
-      </Button>
-    </Teleport>
-  ))
-  TriggerButton.displayName = "TriggerButton"
-
   return (
     <Dialog
       className={clsx("sid-gdpr-consent-dialog", styles.dialog, className)}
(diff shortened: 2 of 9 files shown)


packages/react/src/components/gdpr-consent-dialog/consent-dialog.tsx
```diff
@@ -68,6 +69,25 @@ export const ConsentDialog = ({
     }
   }, [open]);
 
+  const TriggerButton = forwardRef<HTMLButtonElement>((props, ref) => (
```
ahmed-gz, line 72: Wouldn't it be best to pull this out of the `ConsentDialog` component?

packages/react/src/components/teleport/util.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Searches the DOM for a teleport target element
+ * with id [teleportId]. If none is found, creates
+ * one at the root of the DOM.
+ * 
+ * @param teleportKey the key of teleport target
+ * 
+ * @returns DOM selector for teleport target i.e. #foo
+ */
+export const findOrCreateTeleportTarget = (teleportKey: string): HTMLElement => {
+  const element = globalThis.document.getElementById(teleportKey)
```
ahmed-gz, line 13: Wouldn't `teleportElementId` or `teleportTargetId` be better since we use it here as an element ID?

packages/react/src/components/teleport/util.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Companion utility for <Teleport />
+ * 
+ * Searches the DOM for a teleport target element
+ * with id [teleportId]. If none is found, creates
+ * one at the root of the DOM.
+ * 
+ * @param teleportKey the key of teleport target
+ * 
+ * @returns DOM selector for teleport target i.e. #foo
```
ahmed-gz, line 10: I think you meant DOM element instead of selector here?

ahmed-gz (commented):
I tested it locally and it looks good, left some minor comments!

packages/react/src/components/teleport/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+ * @param to The destination portal name
+ * @param children The content to be teleported
+ * @param renderKey (optional) A key to control when the content rerenders
+ */
+export const Teleport = ({ to, children, renderKey }: TeleportProps) => {
+  const target = findOrCreateTeleportTarget(to)
+
+  return (
+    <>
+      {createPortal(children, target, renderKey)}
```
ikovic, line 23: @jake-slashid I was looking at the MUI modal docs and noticed [their warning](https://mui.com/material-ui/react-modal/#server-side-modal) about SSR and `createPortal`.

Our components should support SSR out of the box - anything we can do here to help that?

packages/react/src/components/dialog/index.tsx
```diff
@@ -42,18 +42,33 @@ export const Dialog = ({
...
     <RadixDialog.Root modal={modal} open={open} onOpenChange={onOpenChange}>
       <RadixDialog.Trigger asChild>{trigger}</RadixDialog.Trigger>
-      <RadixDialog.Portal container={container}>
+      <RadixDialog.Portal container={__container}>
         {modal && <RadixDialog.Overlay className={styles.overlay} />}
         <RadixDialog.Content
           className={clsx("sid-dialog", styles.wrapper, className)}
           onInteractOutside={(e) => {
-            if (!dismissable) {
+            if (!modal || !dismissable) {
```
ikovic, line 71: Thinking about it, is there a need to have both `modal` and `dismissable` as props? I understand this makes the API more flexible, but then the `modal` prop loses its meaning.

I would expect the following behaviour:
- `modal: true`: outside interactions are disabled, the user needs to interact with the dialog, close button is rendered
- `modal: false`: modal is closed on interactions outside, close button is not rendered

Maybe the real question is if there is a case when someone wants to explicitly prohibit any interaction with the app until the consent is resolved?

packages/react/src/components/teleport/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+}
+
+/**
+ * Renders child content in a portal element at the DOM root
+ * 
+ * @param to The destination portal name
+ * @param children The content to be teleported
+ * @param renderKey (optional) A key to control when the content rerenders
+ */
+export const Teleport = ({ to, children, renderKey }: TeleportProps) => {
```
ikovic, line 18: Following @ahmed-gz's comments - we should make it clear that `to` is an element ID.

packages/react/src/components/teleport/index.tsx
jake-slashid: I hadn't considered this

packages/react/src/components/teleport/util.ts
jake-slashid: You're right, this is bad copy/paste from something I wrote in the past 🤥

packages/react/src/components/gdpr-consent-dialog/consent-dialog.tsx
jake-slashid: I think as a default probably no 🤔 

If this is its own component it increases the burden of complexity for the minimum implementation

packages/react/src/components/teleport/util.ts
jake-slashid, line 10:
```suggestion
 * @returns the portal element
```

packages/react/src/components/teleport/util.ts
jake-slashid: This feels like leaking an implementation detail, the idea of a "teleport key" is to abstract this detail away.

You choose an identifier and the rest if taken care for you

packages/react/src/components/teleport/index.tsx
jake-slashid: Whats the benefit of this if `<Teleport>` creates the portal element for you?

I don't think I understand why it matters what the selector scheme is

packages/react/src/components/dialog/index.tsx
ikovic: Sorry, when writing this I was convinced I was looking at the `GDPRConsentDialog` and now the lower level dialog component! Agreed with everything you said, thank you!

packages/react/src/components/teleport/index.tsx
ikovic: After taking a look at how you used this component, I think I have a better understanding of the wanted level of abstraction, I was looking at it on a lower level. `to` scheme doesn't really matter here since that's only a name used internally to identify the container. All good, thank you!

packages/react/src/components/teleport/util.ts
ikovic: Based [on this](https://github.com/slashid/javascript/pull/141#discussion_r1308514522) I think we're ok with it as it is.

packages/react/src/components/teleport/index.tsx
jake-slashid: Sorry if that reply came off defensive.

I had a think about this while I was out and a benefit would be that we allow `<Teleport>` to be used with existing elements which weren't created using `<Teleport>`. In theory this kind of API should probably also support passing an element directly.

I'm leaning towards us not supporting this feature though, you could just use `createPortal()` for both of these use cases.

Thought dump:
I see `<Teleport>` as a "managed portal" in concept, so we want to retain control over the portal generation and resolution. For example in future we might change the id generation scheme i.e. `slashid-teleport-${to}` to reduce collisions, or decorate the portal with some css etc.

jake-slashid: Okay all that said I will leave it as is for now, lets revisit this with hindsight once it's in use

packages/react/src/components/teleport/index.tsx
jake-slashid: We might be able to add "support" for SSR in this component behind the abstraction i.e. don't render when SSR detected or perform some alternative rendering operation 🤔

packages/react/src/components/teleport/index.tsx
ikovic: No worries, all good! I wrote the comment before I fully understood the idea behind it - once I saw how you used it in context of `<Dialog>` it clicked with me. Initially I assumed the container would be created externally and this would just lock on it, but it makes perfect sense to consider this as a managed component. I think we should leave it as it is, but maybe add a note in the TS docs for clarity?

packages/react/src/components/teleport/index.tsx
ikovic: We don't have a standard approach yet, definitely something that should be discussed. Maybe we should detect if we're server side and not use portals in that case? That could be handled on a higher level than this.

packages/react/src/components/gdpr-consent-dialog/consent-dialog.tsx
jake-slashid: Done, is it better now?

packages/react/src/components/teleport/index.tsx
jake-slashid: Good idea, done!

packages/react/src/components/dialog/index.tsx
```diff
@@ -42,18 +42,33 @@ export const Dialog = ({
...
+   * Check if the container function return value changed every cycle but only
+   * trigger another render when it actually changes.
+   */
+  const containerFuncReturn = (() => {
+    if (typeof container !== "function") return null
+
+    return container()
+  })()
+
+  const __container = useMemo(() => {
```
ikovic, line 56: @jake-slashid I see the comment but I am not sure if I understand the concern here. Are you making sure that we don't accidentally send a new `container` prop value with each render cycle?

jake-slashid: Yes but I realise now that we could simplify this to:
```tsx
<RadixDialog.Portal
container={
  typeof container === "function"
    ? container()
    : container
}>
```

Because the dom node reference should be stable.

I think at the time for some reason I was concerned about the function reference causing renders.

jake-slashid: Changed, is that better?

ikovic (approved):
Great work @jake-slashid - thank you for paying attention to details!

packages/react/src/components/teleport/index.tsx
jake-slashid: Context in this thread: https://slashidworkspace.slack.com/archives/C04TBH6G59P/p1693311213320299

We're skipping SSR support for now

packages/react/src/components/dialog/index.tsx
ikovic: Yes, thank you!

Issue #142, opened 2023-08-29
Bug: clicking on accept all in the consent dialog doesn't update the toggles in customizing mode


ahmed-gz:
Steps:

- running the dev app locally
- not authenticated
- the dialog opens, I click accept all
- the dialog minimizes
- if I click the icon to bring the dialog up again and then click customize, I expect to see all the options checked
- no options are checked instead

If I reload the page, the state is correct. So I guess once we save the changes on accept all, we don’t update the UI state.

PR #144, opened 2023-08-29
[Issue-142] Fix out of sync consent settings

ahmed-gz:
## Description

[Github issue](https://github.com/slashid/javascript/issues/142)

The consent settings state gets initialized after fetching the consents but it gets out of sync when an action is made from clicking on accept or reject all buttons, this PR fixes the issue by updating the state after a successful action.

## Checklist


packages/react/src/components/gdpr-consent-dialog/state.ts
```diff
@@ -14,10 +16,21 @@ type Action =
...
 export type Dispatch = React.Dispatch<Action>;
 
+export const mapConsentsToSettings = (consents: GDPRConsent[]) => {
+  const settings = Object.fromEntries(
+    CONSENT_LEVELS_WITHOUT_NONE.map((level) => [
+      level,
+      consents.map(({ consent_level }) => consent_level).includes(level),
+    ])
+  );
+  return settings as ConsentSettings;
```
jake-slashid, line 31: How come the cast is necessary here? Casting like this is usually a code smell

packages/react/src/components/gdpr-consent-dialog/actions.tsx
```diff
@@ -48,6 +48,7 @@ export const Actions = ({
     try {
       const consents = await updateGdprConsent(consentLevels);
       onSuccess?.(consents);
+      dispatch({ type: "SET_CONSENT_SETTINGS", payload: consents });
```
jake-slashid, line 51: If there is a refactor on the cards I would like suggest not passing the `dispatch()` function reference into child components. This creates an unnecessary coupling, but my main issue is the added complexity this pattern brings.

It's quite difficult to reason about and took me some time to get my head around.

My suggestion: `<Actions>` should be presentational only, use "events" to signal user interaction and update state.

```ts
<Actions
  onAcceptAll={dispatch({ "ACCEPT_ALL", payload: defaultAcceptAllLevels })}
  onRejectAll={...}
  onCustomize={...}
  loading={...}
/>
```

But as I start to unwind the component like this, it doesn't feel like the abstraction adds very much value - it's mostly indirection.

Why don't we just move the four buttons into `<ConsentDialog>`?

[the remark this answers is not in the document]
jake-slashid: My other comment has some very high level examples of how we might group actions into "events" conceptually

I'd like to ask: do we actually need `useReducer()` here?

It adds a lot of overhead and the state is only being referenced in two places, which could be presentational components: `<Actions>`, `<Settings>`

packages/react/src/components/gdpr-consent-dialog/state.ts
ahmed-gz: `Object.fromEntries` here doesn't preserve the key's type and simply returns a key as string so a cast is necessary, do you know a way to handle this without casting?

.changeset/breezy-pillows-kick.md
```diff
@@ -0,0 +1,5 @@
+---
+"@slashid/react": minor
```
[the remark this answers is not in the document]
ahmed-gz: you're right, sorry I missed that!

packages/react/src/components/gdpr-consent-dialog/state.ts
jake-slashid: Ah I see, okay

There's two things we would need to do to remove the cast:
- Type the map correctly so it's clear the return type is a tuple
- Create a typed wrapper for `Object.fromEntries`

// util.ts
export const fromEntries = <const T extends ReadonlyArray<readonly [PropertyKey, unknown]>>(
  entries: T
): { [K in T[number] as K[0]]: K[1] } => {
  return Object.fromEntries(entries) as { [K in T[number] as K[0]]: K[1] };
};
```

```
// state.ts
import { fromEntries } from './util.ts'

const settings: ConsentSettings = fromEntries(
  CONSENT_LEVELS_WITHOUT_NONE.map<[typeof CONSENT_LEVELS_WITHOUT_NONE[number], boolean]>((level) => [
    level,
    consents.map(({ consent_level }) => consent_level).includes(level),
  ])
)
```

Now we have type safety

packages/react/src/components/gdpr-consent-dialog/actions.tsx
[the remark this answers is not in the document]
ahmed-gz: I usually go for `useReducer()` when I need more than 3 state variables, in `<ConsentDialog>` we need 5 state variables that also need to be shared with child components: `<Actions>` and `<Settings>`, so IMO `useReducer()` feels cleaner than using `useState()` variables and also allows streamlining state updates. Does that make sense?

ahmed-gz: I agree with your point about passing `dispatch`, I will change it.
Regarding moving the buttons, I just feel like `<ConsentDialog>` will get bloated too much if it has those buttons, but if you feel like it will improve readability I will move them.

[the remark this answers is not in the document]
jake-slashid: Okay! I can see how a reducer can make the state more predictable here, I think it will probably make sense to me after refactoring the actions.

jake-slashid (approved):
Merging `SET_CONSENT_SETTINGS` and `SET_OPEN` is a clear improvement 🙏 

I think the component overall is a bit easier to reason about

ikovic (approved):

kaspermroz (approved):

PR #145, opened 2023-08-29
[SID-1214] Generate npmrc before publish

jake-slashid:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1214)

Unstable publish [failed](https://github.com/slashid/javascript/actions/runs/6013643387/job/16311531987) because of missing npm token

## Checklist

### DIFF
--- a/.github/workflows/unstable-release.yml
+++ b/.github/workflows/unstable-release.yml
@@ -1,4 +1,4 @@
-name: Unstable/nightly release
+name: Unstable/next release
 
 on:
   push:
@@ -6,7 +6,7 @@ on:
 
 jobs:
   unstable-release:
-    name: Unstable/nightly release
+    name: Unstable/next release
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


ikovic (approved):

.github/workflows/unstable-release.yml
```diff
@@ -30,6 +30,9 @@ jobs:
         
       - name: Build packages
         run: pnpm build
-        
+
+      - name: Setup npmrc
```
ikovic, line 34: Can you use the secrets as we do for the regular release flow?

```
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```

jake-slashid: Does this work? `NPM_TOKEN=... npm publish`

Because that's how I understand standard publishing to work - afaik you need to have an `.npmrc` like I've created here.

The difference between this and the example you're referencing is that it uses `changesets/action` which I assume has support for passing `NPM_TOKEN` as env var?

https://github.com/changesets/action#with-publishing

Issue #139, opened 2023-08-28
Implement the OTP code input to match the design


ahmed-gz:
Implement the OTP code input to match the design here: https://www.figma.com/file/nOXz0jhwoHGGXJwOdfTmsa/SlashID-%E2%80%94-Sign-in%2FSign-up-Design-by-Significa?type=design&node-id=204-3659&mode=design&t=C5ZZZSbHtu2eWbuM-0

PR #152, opened 2023-09-05
[Issue-139] Implement the OTP code input to match the design

ahmed-gz:
## Description

[Github issue](https://github.com/slashid/javascript/issues/139)

- Replaced the native select box with our existing `<Dropdown>` component to match the design.
- Fixed the emoji flags not showing on windows using [country-flag-emoji-polyfill](https://github.com/talkjs/country-flag-emoji-polyfill).
- Implemented the  `<OtpInput>` component.
- Improved the OTP authenticating states (updated text and icons, and hid the retry section when submitting).
- Added [`vitest-canvas-mock`](https://github.com/wobsoriano/vitest-canvas-mock) dependency to fix an error in the unit tests coming from [country-flag-emoji-polyfill](https://github.com/talkjs/country-flag-emoji-polyfill) related to `HTMLCanvasElement.prototype.getContext`:
image

## Checklist


ikovic:
@ahmed-gz this change somehow breaks the static build of the demo nextJS application - there should be no references to `window` in the first render. I don't see any direct references to `window` in the changes you submitted, so it's possible that the library update messed this up. Can you investigate where exactly this happens please?

ahmed-gz:
> @ahmed-gz this change somehow breaks the static build of the demo nextJS application - there should be no references to `window` in the first render. I don't see any direct references to `window` in the changes you submitted, so it's possible that the library update messed this up. Can you investigate where exactly this happens please?

@ikovic thanks for checking that it's exactly like you said, the new version of the library has a reference to `window` in its source code here:
image

I tried wrapping the `CountryList` class with a hook that uses `useEffect` but it seems that this error gets triggered once we import the library, I think I will have to revert to the old version and find another way to show the flags

ikovic:
@ahmed-gz if it's a side effect of importing the library, then it can probably be fixed by rendering the parent component client side only. We can do this in a nice way by having a `ClientSide` wrapper component that only renders children client side and wrapping the minimal component with this. Eg we should extract the code that imports the library in a module with the smallest possible surface and only render it client side.

ahmed-gz:
> @ahmed-gz if it's a side effect of importing the library, then it can probably be fixed by rendering the parent component client side only. We can do this in a nice way by having a `ClientSide` wrapper component that only renders children client side and wrapping the minimal component with this. Eg we should extract the code that imports the library in a module with the smallest possible surface and only render it client side.

@ikovic I just tested the Country Flag Emoji Polyfill and it works as expected in next and remix, please let me know if you see any issues with this approach.

ikovic:
@ahmed-gz thanks, looks good to me! Tested it locally, works as expected. Also the bundle size does not seem to be impacted much.

ahmed-gz:
> working

Thanks a lot for testing so thoroughly 🙏

ikovic:
@ahmed-gz we'll leave this one open and merge it early next week - we're doing a release now that won't include this change

packages/react/package.json
```diff
@@ -57,6 +57,7 @@
     "@vanilla-extract/recipes": "^0.3.0",
     "@vanilla-extract/sprinkles": "^1.5.1",
     "compress.js": "^1.2.2",
+    "country-flag-emoji-polyfill": "^0.1.4",
```
ikovic, line 60: @ahmed-gz this is no longer used, correct?

packages/react/package.json
```diff
@@ -92,6 +93,7 @@
     "tsconfig": "workspace:*",
     "vite": "^3.1.0",
     "vitest": "^0.25.2",
+    "vitest-canvas-mock": "^0.3.3",
```
ikovic, line 96: Do we still need this?

ikovic (approved):

packages/react/package.json
ahmed-gz: @ikovic it's still used actually to show the flag for the selected country in the form

packages/react/package.json
ahmed-gz: yes, it's still needed to fix an error in the unit tests coming from [country-flag-emoji-polyfill](https://github.com/talkjs/country-flag-emoji-polyfill) related to `HTMLCanvasElement.prototype.getContext`:
image

packages/react/src/components/dropdown/index.tsx
```diff
@@ -1,13 +1,14 @@
...
+import { ReactNode, useCallback, useState } from "react";
 import { Check } from "../icon/check";
 import { ChevronDown } from "../icon/chevron-down";
 import * as styles from "./dropdown.css";
-import { ReactNode, useCallback, useState } from "react";
 
 type Item = {
   label: ReactNode;
   value: string;
+  textValue?: string;
```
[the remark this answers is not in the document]
ahmed-gz: @ikovic I will open a PR for this, `textValue` is used here for typeahead purposes for complex content, for example when you need to add an icon at the beginning of the text which will prevent the typeahead to work correctly, when we pass the searchable content without the icon to `textValue` we can control the typeahead behavior without interfering with the displayed text. Whereas `label` is simply the text rendered in the item text.
Does it make sense?

PR #155, opened 2023-09-12
[SID-1236] Slots API

ikovic:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1236)

Introducing the Slots API that enables customers to extend the `<Form>` by composition.

### TODO

- [ ] Potentially merge the two context providers
- [ ] Refactor the `<DynamicFlow>` to use a composed form
- [x] Write tests that confirm the extended components are rendered
- [x] Confirm that the module resolution issues are fixed (Vite has trouble importing nested modules)
- [x] Complete the API: `authenticating`, `error` and `success` slots
- [x] Add the demo-form example for a composed API
- [x] Add storybook entries
- [x] Write docs
- [x] Make sure SSR works - no hydration errors or warnings

## Checklist

### DIFF
--- a/packages/react/src/components/dynamic-flow/configured-handle-form.tsx
+++ b/packages/react/src/components/dynamic-flow/configured-handle-form.tsx
@@ -1,9 +1,9 @@
 import { Factor } from "@slashid/slashid";
-import { Handle, HandleType } from "../../../domain/types";
-import { useConfiguration } from "../../../hooks/use-configuration";
-import { sprinkles } from "../../../theme/sprinkles.css";
-import { Divider } from "../../divider";
-import { Tabs } from "../../tabs";
+import { Handle, HandleType } from "../../domain/types";
+import { useConfiguration } from "../../hooks/use-configuration";
+import { sprinkles } from "../../theme/sprinkles.css";
+import { Divider } from "../divider";
+import { Tabs } from "../tabs";
 import { HandleForm } from "./handle-form";
 import { useMemo } from "react";
 import {
@@ -11,7 +11,7 @@ import {
   hasOidcAndNonOidcFactors,
   isFactorOidc,
   resolveLastHandleValue,
-} from "../../../domain/handles";
+} from "../../domain/handles";
 
 export const TAB_NAME = {
   email: "email",
--- a/packages/react/src/components/dynamic-flow/handle-form.tsx
+++ b/packages/react/src/components/dynamic-flow/handle-form.tsx
@@ -5,19 +5,18 @@ import {
   filterFactors,
   isFactorOidc,
   parsePhoneNumber,
-} from "../../../domain/handles";
-import { HandleType, Handle } from "../../../domain/types";
-import { useConfiguration } from "../../../hooks/use-configuration";
-import { useForm } from "../../../hooks/use-form";
+} from "../../domain/handles";
+import { HandleType, Handle } from "../../domain/types";
+import { useConfiguration } from "../../hooks/use-configuration";
+import { useForm } from "../../hooks/use-form";
 
-import { Button } from "../../button";
-import { Dropdown } from "../../dropdown";
-import { Flag, GB_FLAG, PhoneInput, Input } from "../../input";
-import { ErrorMessage } from "../error-message";
-import { isValidPhoneNumber, isValidEmail } from "../validation";
-import { TextConfigKey } from "../../text/constants";
-import { useSlots, type Props as SlotProps } from "../../slot";
-import { sprinkles } from "../../../theme/sprinkles.css";
+import { Button } from "../button";
+import { Dropdown } from "../dropdown";
+import { Flag, GB_FLAG, PhoneInput, Input } from "../input";
+import { ErrorMessage } from "../form/error-message";
+import { isValidPhoneNumber, isValidEmail } from "../form/validation";
+import { TextConfigKey } from "../text/constants";
+import { sprinkles } from "../../theme/sprinkles.css";
 import React from "react";
 
 export const FACTOR_LABEL_MAP: Record<
@@ -38,7 +37,6 @@ export type Props = {
   handleSubmit: (factor: Factor, handle: Handle) => void;
   validate?: Validator<string>;
   defaultValue?: string;
-  children?: React.ReactElement<SlotProps<"initial.submit">>[];
 };
 
 export const HandleForm: React.FC<Props> = ({
@@ -46,7 +44,6 @@ export const HandleForm: React.FC<Props> = ({
   factors,
   handleSubmit,
   defaultValue,
-  children,
 }) => {
   const filteredFactors = useMemo(
     () => filterFactors(factors, handleType).filter((f) => !isFactorOidc(f)),
@@ -61,24 +58,6 @@ export const HandleForm: React.FC<Props> = ({
   );
   const [factor, setFactor] = useState<Factor>(filteredFactors[0]);
   const { text } = useConfiguration();
-
-  const defaultSlots = React.useMemo(() => {
-    return {
-      "initial.submit": (
-        <Button
-          className={sprinkles({ marginTop: "6" })}
-          type="submit"
-          variant="primary"
-          testId="sid-form-initial-submit-button"
-          disabled={status === "invalid"}
-        >
-          {text["initial.submit"]}
-        </Button>
-      ),
-    };
-  }, [status, text]);
-
-  const slots = useSlots({ children, defaultSlots });
 
   useEffect(() => {
     setFactor(filteredFactors[0]);
@@ -170,7 +149,15 @@ export const HandleForm: React.FC<Props> = ({
       )}
       {input}
       <ErrorMessage name={handleType} />
-      {slots["initial.submit"]}
+      <Button
+        className={sprinkles({ marginTop: "6" })}
+        type="submit"
+        variant="primary"
+        testId="sid-form-initial-submit-button"
+        disabled={status === "invalid"}
+      >
+        {text["initial.submit"]}
+      </Button>
     </form>
   );
 };
--- a/packages/react/src/components/form/form-customisation.test.tsx
+++ b/packages/react/src/components/form/form-customisation.test.tsx
@@ -4,6 +4,7 @@ import { TestSlashIDProvider } from "../../context/test-slash-id-provider";
 import { createTestUser, inputEmail } from "../test-utils";
 import { Slot } from "../slot";
 import userEvent from "@testing-library/user-event";
+import { ConfigurationProvider } from "../../main";
 
 describe("#Form - customisation", () => {
   test("should render the footer slot", () => {
@@ -63,7 +64,7 @@ describe("#Form - customisation", () => {
     expect(screen.findByTestId("sid-form-authenticating-state")).toBeFalsy();
   });
 
-  test.only("should render the error slot", async () => {
+  test("should render the error slot", async () => {
     const logInMock = vi.fn(() => Promise.reject("login error"));
     const user = userEvent.setup();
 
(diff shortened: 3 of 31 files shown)


packages/react/src/components/dynamic-flow/configured-handle-form.tsx
```diff
@@ -1,24 +1,24 @@
 import { Factor } from "@slashid/slashid";
```
ikovic, line 1: Moved this file to `dynamic-flow` - the scope of this ticket blew up, so I had to make a cut here. A follow up task would be to reimplement the DynamicFlow internally using the slots/composition API.

packages/react/src/components/dynamic-flow/handle-form.tsx
```diff
@@ -5,18 +5,19 @@ import {
   filterFactors,
```
ikovic, line 5: Explained [here](https://github.com/slashid/javascript/pull/155/files#r1324416382).

packages/react/src/components/form/form.tsx
```diff
@@ -14,42 +14,127 @@ import {
...
+
+export type InternalFormContextType = {
+  flowState: ReturnType<typeof useFlowState> | null;
+  lastHandle?: Handle;
+  submitPayloadRef: React.MutableRefObject<PayloadOptions>;
+  handleSubmit: (factor: Factor, handle?: Handle) => void;
+  selectedFactor?: Factor;
+  setSelectedFactor: React.Dispatch<React.SetStateAction<Factor | undefined>>;
+};
+export const InternalFormContext = React.createContext<InternalFormContextType>(
```
ikovic, line 44: Context is used to share values between nested components.

packages/react/src/components/form/initial/controls.tsx
```diff
@@ -0,0 +1,407 @@
...
+        text,
+        handleSubmit,
+      }: ControlsProps) => React.ReactNode);
+};
+
+/**
+ * Component responsible for rendering the form and the related controls.
+ * Sets up the form context and provides the submit handler.
+ */
+export const Controls = ({ children }: Props) => {
```
ikovic, line 70: This module has some duplication - I can do another pass refactoring this out. If you guys have any good ideas on how to structure this differently please let me know.

packages/react/src/components/slot/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+// TODO consider using asChild prop as an alternative to render props = passes props to the immediate child
+
+// add render props here
+
+/**
+ * TODO document this component
+ * - issue: it is not typesafe, it is possible to render a slot with a name that does not exist
+ * @returns
+ */
+export function Slot<N extends string>({ children, name, asChild }: Props<N>) {
```
ikovic, line 18: I couldn't get TS to help - I wanted to see if we could somehow annotate the components with slots on the type level, so TS gives you instant feedback when passing down a slot with an incorrect name.

packages/react/src/components/slot/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+  if (asChild) {
```
ikovic, line 19: TODO out of scope for now IMO, I can add this in a follow up PR - don't want to create a large API immediately, got to be sure the design is sound.

packages/react/src/components/form/form.tsx
jake-slashid: Does this work with SSR? I guess only the initial state matters for SSR right?

packages/react/src/components/form/form-customisation.test.tsx
```diff
@@ -0,0 +1,248 @@
...
+import { createTestUser, inputEmail } from "../test-utils";
+import { Slot } from "../slot";
+import userEvent from "@testing-library/user-event";
+import { ConfigurationProvider } from "../../main";
+import { useState } from "react";
+import { Factor } from "@slashid/slashid";
+import { Handle } from "../../domain/types";
+
+describe("#Form - customisation", () => {
+  test("should render the footer slot", () => {
```
kaspermroz, line 13: As per README.md:
 > In addition to that, any components passed in without a wrapper will be ignored in the top level containers.

I'd add a small test case for that

packages/react/src/components/form/form.tsx
kaspermroz: Didn't we drop the Context option because of SSR issues? @ikovic did you manage to find a workaround for that?

packages/react/src/components/form/form.tsx
```diff
@@ -14,42 +14,127 @@ import {
...
+    flag: undefined,
+  });
+  const [selectedFactor, setSelectedFactor] = React.useState<
+    Factor | undefined
+  >();
+  const { status } = flowState;
+
+  const defaultSlots = React.useMemo(() => {
+    const slots = {
+      footer: showBanner ? <Footer /> : null,
```
kaspermroz, line 81: I'm thinking if we could simply drop `showBanner` now that we have slots API

packages/react/src/components/form/initial/controls.tsx
kaspermroz: I'd think of nesting this in `/controls/<component>.tsx` and exporting Controls from `index`., as well as
```
Controls.displayName = "Controls";
Controls.Input = FormInput;
Controls.Submit = Submit;
```

ikovic:
@kaspermroz @jake-slashid there were two questions re: SSR and the Context API - on its own the Context API does not clash with SSR, the only problem is when the initial render is empty because the context values get initialised after the first render. The use case we discussed previously that does not work with SSR is this:
- parent must render fallback components if the developer does not provide their own components
- these components register themselves after the first render using the context
- this way initially the parent renders empty, no children have had a chance to register themselves yet

In this case we're just using the Context API to decouple children from parents and let us pass values to deeply nested components.

I'll confirm if SSR works OK once I add this to the demo app, but I believe it should be fine 😄

packages/react/src/components/form/form-customisation.test.tsx
ikovic: Agreed, will add immediately!

packages/react/src/components/form/form.tsx
ikovic: Agreed, it would be a breaking change though. We could deprecate it and work around it like this:
- mark as deprecated, but if used it just sends an empty slot there under the hood
- in the next major version we drop it from the API entirely

ikovic:
@kaspermroz:

> I'd think of nesting this in /controls/<component>.tsx and exporting Controls from index., as well as...

Vite has some trouble with nesting like this, I am still not 100% sure why. We might be creating some circular dependencies this way. I do agree this could be done in a tidier way so given that this is working, my plan was to sort out any bug fixes and PR comments first and then do this as the last pass. Sounds ok?

jake-slashid (approved):
Being such a big refactor I'm avoiding nit picks in this PR, for me what's most important is getting the API right and getting it delivered on time.

That said - I think the API looks good and internally I really like the `useSlots()` helper, this is the kind of internal API I was hoping you'd make so it's easy to introduce slots elsewhere!

If you have a particular area you feel is weak and want a second opinion please point that out specifically and I'll give it some focus.

packages/react/src/components/slot/index.tsx
```diff
@@ -0,0 +1,52 @@
+import React from "react";
+import { MaybeArray } from "../../domain/types";
+
+export type Props<Name extends string> = {
+  children: React.ReactNode;
+  name: Name; // name of the slot the component is being rendered into
+};
+
+export type Slots<SlotNames extends string> = MaybeArray<
```
ikovic, line 9: Unfortunately at the moment TypeScript cannot test the children types:
- https://www.arahansen.com/how-children-types-work-in-react-18-and-typescript-4/
- https://stackoverflow.com/a/71987357
- https://github.com/microsoft/TypeScript/issues/21699

I still wanted to add this type in case it changes in the future - for the moment we can at least use it as docs, even though it is not enforced.

packages/react/src/components/form/form-customisation.test.tsx
ikovic: Done!

packages/react/src/components/form/form.tsx
```diff
@@ -14,42 +14,134 @@ import {
...
+import { Factor } from "@slashid/slashid";
+import { Flag } from "../input";
 
 export type Props = ConfigurationOverridesProps & {
   className?: string;
   onSuccess?: CreateFlowOptions["onSuccess"];
   middleware?: LoginOptions["middleware"];
+  children?: Slots<
+    "initial" | "authenticating" | "success" | "error" | "footer"
+  >; // TS does not enforce this, but it is used for documentation
```
ahmed-gz, line 29: @ikovic if you think it could be helpful, we could at least enforce the slot name by exporting a specific `<FormSlot>` component instead of the generic `<Slot>` that will enforce the name type like this:
```
type FormSlotName = "initial" | "authenticating" | "success" | "error" | "footer";

export function FormSlot({ children, name }: SlotProps<FormSlotName>) {
  return <Slot name={name}>{children}</Slot>;
}
```

packages/react/src/components/form/form.tsx
```diff
@@ -14,42 +14,134 @@ import {
...
+        <ConfigurationOverrides text={text} factors={factors}>
+          {flowState.status === "initial" && (
+            <FormProvider>{slots.initial}</FormProvider>
+          )}
+          {flowState.status === "authenticating" && (
+            <FormProvider>{slots.authenticating}</FormProvider>
+          )}
+          {flowState.status === "success" && slots.success}
+          {flowState.status === "error" && slots.error}
+          {slots.footer}
```
ahmed-gz, line 140: was rendering this regardless for `showBanner` value intentional?

packages/react/src/components/form/initial/header.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  const Heading = React.useMemo(() => {
+    return (
+      <div className={styles.header}>
+        <Text
+          as="h1"
+          variant={{ size: "2xl-title", weight: "bold" }}
+          t="initial.title"
+        />
+        <Text variant={{ color: "tertiary" }} as="h2" t="initial.subtitle" />
```
ahmed-gz, line 26: @ikovic Since the components were broken down, there is a change I made in this [PR](https://github.com/slashid/javascript/pull/156) which won't be reflected here, so could you please change the `variant` to this: `variant={{ color: "contrast", weight: "semibold" }}`

ahmed-gz (commented):
@ikovic This looks really good awesome job 👍 I just had few comments!

packages/react/src/components/slot/index.tsx
ahmed-gz: I think we could enforce the type on the name at least with this suggestion: https://github.com/slashid/javascript/pull/155#discussion_r1325874300

packages/react/src/components/form/form.tsx
ikovic: It's already handled on top, when constructing the default slots.

packages/react/src/components/form/form.tsx
ikovic: This would break the implementation - only a `Slot` instance can be rendered as an immediate child of the parent container.

jake-slashid: I feel quite strongly that we should keep `<Slot>` as the top-level API so we can build some raport with consumers, and grow their understanding of this concept.

That said maybe a compromise is to offer the type as an export to use as a generic? It's ugly but some might find it helpful

```ts
const Component = () => {
  return (
    <Slot<FormSlots> name="initial">
    </Slot>
  )
}
```

PR #164, opened 2023-09-27
Improve the OTP flow states

ahmed-gz:
## Description

## Checklist

### DIFF
--- a/.changeset/olive-onions-prove.md
+++ b/.changeset/olive-onions-prove.md
@@ -2,4 +2,5 @@
 "@slashid/react": patch
 ---
 
-Improve the OTP flow states
+Update text and icons for the OTP authenticating states
+Hide the retry section when the OTP code is being submitted
--- a/.changeset/olive-onions-prove.md
+++ b/.changeset/olive-onions-prove.md
@@ -1,0 +1,5 @@
+---
+"@slashid/react": patch
+---
+
+Improve the OTP flow states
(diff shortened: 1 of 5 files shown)


.changeset/olive-onions-prove.md
```diff
@@ -0,0 +1,5 @@
+---
+"@slashid/react": patch
+---
+
+Improve the OTP flow states
```
ikovic, line 5: This message will end up in the automatically generated changelog - we need to be a bit more specific than this. Can we list the changes we did here?

packages/react/src/domain/handles.ts
```diff
@@ -112,7 +122,8 @@ export type AuthenticatingMessage = {
   message: TextConfigKey;
 };
 export function getAuthenticatingMessage(
```
ikovic, line 124: Can we check if this function is being used anywhere else but in `authenticating.tsx`? If not, I would suggest moving it to the same file so it has a stronger context.

ahmed-gz: You're right, it's only used in in `authenticating.tsx` I will move it there

.changeset/olive-onions-prove.md
ahmed-gz: Sorry you're right, I will add more details

ikovic (approved):
Thanks @ahmed-gz, this looks good!

PR #169, opened 2023-10-12
[SID-1323] Add last handle write test

kaspermroz:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1323)

Following the fix for last handle, we want to extend the tests to cover the blind spots.

## Checklist

### DIFF
--- a/packages/react/src/components/test-utils.ts
+++ b/packages/react/src/components/test-utils.ts
@@ -3,6 +3,12 @@ import { TEXT } from "./text/constants";
 import { FactorMethod, OrganizationDetails, User } from "@slashid/slashid";
 import { faker } from "@faker-js/faker";
 import { Handle } from "../domain/types";
+import {
+  Handler,
+  PublicReadEvents,
+  SlashID,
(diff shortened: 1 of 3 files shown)


packages/react/src/components/test-utils.ts
```diff
@@ -119,3 +125,55 @@ export const polyfillPointerEvent = () => {
   window.HTMLElement.prototype.hasPointerCapture = vi.fn();
   window.HTMLElement.prototype.releasePointerCapture = vi.fn();
 };
+
+/**
+ * This class is a wrapper on top of the SlashID instance, with mocked
+ * pubsub mechanism. This allows us to test all subscription-related logic
+ * with ease, without the need of accessing private SlashID emitter property.
+ */
+export class MockSlashID extends SlashID {
```
jake-slashid, line 134: Extending the class is a good solution!

jake-slashid (approved):

ikovic (approved):

PR #182, opened 2023-11-15
[SID-1368] Record logout and person identified analytics events, make analytics opt-out

jake-slashid:
Dependencies:
- https://github.com/slashid/ng-evangelion/pull/1200
- https://github.com/slashid/ng-evangelion/pull/1204

## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1368)

- Any time the user is stored record this as a "person identified" (login, parse from token storage)
- When the user explicitly logs out, record this as a "person logged out"
- Make analytics on by default

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -62,7 +62,7 @@
   },
   "devDependencies": {
     "@faker-js/faker": "^8.0.2",
-    "@slashid/slashid": "3.15.0",
+    "@slashid/slashid": "file:.yalc/@slashid/slashid",
     "@storybook/addon-essentials": "7.0.0-rc.2",
     "@storybook/addon-interactions": "7.4.0",
     "@storybook/addon-links": "7.4.0",
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -62,7 +62,7 @@
   },
   "devDependencies": {
     "@faker-js/faker": "^8.0.2",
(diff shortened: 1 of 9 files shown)


packages/react/package.json
```diff
@@ -62,7 +62,7 @@
   },
   "devDependencies": {
     "@faker-js/faker": "^8.0.2",
-    "@slashid/slashid": "3.15.0",
+    "@slashid/slashid": "file:.yalc/@slashid/slashid",
```
paulo-raca, line 65: How does it work? The version is referenced in a gitignored file? :thinking:

paulo-raca (approved):

packages/react/package.json
jake-slashid: This is just a placeholder, but it's a tool for publishing packages on local machine for development

I'll change this to the actual version once its published

https://github.com/wclr/yalc

jake-slashid: done

packages/react/package.json
```diff
@@ -62,7 +62,7 @@
   },
   "devDependencies": {
     "@faker-js/faker": "^8.0.2",
-    "@slashid/slashid": "3.15.0",
+    "@slashid/slashid": "^3.16.0",
```
ikovic, line 65: Should we go with a fixed version? Just to prevent surprises from happening 😅

jake-slashid: Done!

PR #183, opened 2023-11-20
[SID-1384] Jump page implementation

ikovic:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1384)

Implement the jump page as a React app, using the `@slashid/react-primitives`. Scope of this was to recreate the jump page content as it is, without adding new content.

Password flow will be built in a separate PR.

## Checklist


apps/jump-page/src/layouts/page.astro
```diff
@@ -0,0 +1,33 @@
+---
+interface Props {
+	title: string;
+}
+
+const { title } = Astro.props;
+---
+
+<!DOCTYPE html>
+<html ⚡4email data-css-strict>
```
jake-slashid, line 10: ⚡⚡⚡

Is this a typo?

ikovic: This was present in the markup - I believe whoever implemented this originally took a template that was supposed to be an email 🤔

packages/react-primitives/README.md
```diff
@@ -3,28 +3,78 @@
...
+```jsx
+import { Text as BaseText, TextProps } from "@slashid/react-primitives";
+import { TextConfig } from "./text";
 
-### Text
+type Props = TextProps<TextConfig>;
 
-Text primitives are implemented here, translations are implemented in the consuming app.
-Wrap the text primitives in a `Text` component that will handle the translation.
+export const Text: React.FC<Props> = (props) => {
```
ikovic, line 77: Replace this with a factory function like
```
const { Text, TextProvider } = createText(config)
```

apps/jump-page/src/components/app/app.component.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { useEffect, useMemo, useRef, useState } from "react";
+import { I18N, detectLanguage } from "../../domain/i18n";
+import { Flow, Error } from "../flow";
+import { Card } from "../card";
+
+import * as styles from "./app.css";
+import { config } from "../../config";
+
+export function App() {
+  const isBrowser = !!globalThis.window;
```
kaspermroz, line 21: @ikovic we have something like this in other package, IDK if it makes much difference in that case
```jsx
export function isBrowser(): boolean {
  return typeof window !== "undefined" && globalThis === window;
}
```

kaspermroz (approved):

apps/jump-page/src/components/app/app.component.tsx
ikovic: Yep, this one is available from the react primitives package. I'll see if I can use it instead of reimplementing it here. Thanks!

apps/jump-page/src/layouts/page.astro
ikovic: Fixed - removed this as it is not needed.

apps/jump-page/src/components/app/app.component.tsx
ikovic: Resolved!

packages/react-primitives/README.md
ikovic: @kaspermroz @jake-slashid after a couple of iterations I decided to leave this as it is following this reasoning:
- `textConfig` is a value, not a type
- the keys must remain the same, but the values in this object can/are likely to change (eg in the jump page we have i18n)

So passing the `textConfig` to the factory function implies you cannot change these values during runtime, which is incorrect.

The only drawback of the current approach is having to wrap around the base `<Text>` component, which IMO is acceptable given the fact that this is still internal and the change was already made for the `@slashid/react` package.

jake-slashid (approved):

PR #186, opened 2023-12-06
[SID-1432] Catch react client-side errors

jake-slashid:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1432)

Companion for: https://github.com/slashid/ng-evangelion/pull/1281

We assume any error within `<Form>` which is not a `ResponseError` adds friction to authentication, and report it as a `client_side_error` failure.


jake-slashid:
This needs deduping before we can merge it

PR #194, opened 2023-12-21
[SID-1460] React SDK: do not expose the internal package

ikovic:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1460)

Previously to this PR, `@slashid/react-primitives` was an internal package that had to be exposed via `npm` in order for anyone to install the `@slashid/react` package, given that even though it was internal we still had a reference to it in `package.json`.

With these changes, we remove the dependency and include an alias in order for `@slashid/react` to import the code outside of its relative project root. This way, there is no dependency so customers now only need to install the base package and `@slashid/react-primitives` does not need to be published.

## Checklist

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -22,13 +22,16 @@
     "jwt"
   ],
   "type": "module",
-  "types": "dist/main.d.ts",
+  "types": "dist/react/src/main.d.ts",
   "main": "dist/main.js",
   "module": "dist/main.js",
   "sideEffects": false,
   "exports": {
     ".": {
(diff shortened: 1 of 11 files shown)


packages/react/package.json
```diff
@@ -22,13 +22,16 @@
     "jwt"
   ],
   "type": "module",
-  "types": "dist/main.d.ts",
```
ikovic, line 25: Given that two projects are now being used in a single build process, `tsc` outputs two folders with type declarations in `dist`.

jake-slashid (approved):

PR #209, opened 2024-01-19
[SID-1497] Password support

ikovic:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1497)

This PR removes the `jump-page` app as it was moved to `ng-evangelion`. The main reason was to simplify deployments and version management, since the `jump-page` depends directly on the code SDK and needs to be released at the same time.

To do this, the `react-primitives` package was once again exposed through `npm`, but in a specific way:
- source files are being kept in the package
- only the type declarations are built

This way the consumer of the package can build the app autonomously, which is important because of the way we use `vanilla-extract`.

The biggest TODO left is now writing automated tests.

## Testing guide

- first make sure you check out the branch `sid-1437-password-reset` in the `ng-evangelion` repo
- run `task up` there
- verify that you are targeting the local API in `javascript/packages/react`
- in `javascript` run `pnpm dev --filter react`
- test the password flow in the top left form

## Checklist


packages/react/src/components/form/authenticating/messages.ts
```diff
@@ -0,0 +1,56 @@
+import { Factor } from "@slashid/slashid";
+import { TextConfigKey } from "../../text/constants";
+
+// TODO add case for password
```
kaspermroz, line 4: Did you miss a spot here?

PR #215, opened 2024-02-07
[SID-1542] React SDK: Trigger a render after performing MFA

ikovic:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1542)

After MFA is performed, the user instance is mutated but it does not change identity, skipping the render cycle. In this PR I made sure a new instance is created instead, so the components can react to it as expected.

## Checklist

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@slashid/react",
-  "version": "1.18.1-beta.1",
+  "version": "1.18.0",
   "private": false,
   "publishConfig": {
     "access": "public"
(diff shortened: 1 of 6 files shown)


packages/react/package.json
```diff
@@ -65,7 +65,7 @@
   },
   "devDependencies": {
     "@faker-js/faker": "^8.0.2",
-    "@slashid/slashid": "3.18.0",
+    "@slashid/slashid": "3.18.2-beta.0",
```
ikovic, line 68: Not to be merged until the core SDK is released.

kaspermroz (approved):

PR #216, opened 2024-02-09
[SID-1546] Add possibility to specify allowed handle types for factors with multiple handle types

kaspermroz:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1546)

This change adds `allowedHandleTypes` to `password` factor method, as it's currently the only factor that supports both email and phone number handle types.

The motivation behind this change is: there are valid use cases where the user wants to explicitly declare the supported handle types. Good example is Shopify and SlashID Login app - Multipass works only with emails, so the phone number option should not be shown.

## Checklist

### DIFF
--- a/packages/react-primitives/src/components/tabs/index.tsx
+++ b/packages/react-primitives/src/components/tabs/index.tsx
@@ -12,16 +12,22 @@ type Props = {
   tabs: Tab[];
   className?: string;
   defaultValue?: string;
+  testId?: string;
 };
 
-export const Tabs: React.FC<Props> = ({ className, tabs, defaultValue }) => {
+export const Tabs: React.FC<Props> = ({
+  className,
+  tabs,
+  defaultValue,
+  testId,
+}) => {
   if (!tabs.length) {
     return null;
   }
 
   return (
     <RadixTabs.Root
-      data-testid="sid-handle-type-tabs"
+      data-testid={testId}
       className={clsx("sid-tabs", className)}
       defaultValue={defaultValue ?? tabs[0]?.id}
     >
--- a/packages/react-primitives/src/components/tabs/index.tsx
+++ b/packages/react-primitives/src/components/tabs/index.tsx
@@ -21,6 +21,7 @@ export const Tabs: React.FC<Props> = ({ className, tabs, defaultValue }) => {
 
   return (
     <RadixTabs.Root
+      data-testid="sid-handle-type-tabs"
       className={clsx("sid-tabs", className)}
       defaultValue={defaultValue ?? tabs[0]?.id}
     >
--- a/packages/react/src/components/text/constants.ts
+++ b/packages/react/src/components/text/constants.ts
@@ -11,7 +11,7 @@ export const TEXT = {
   "initial.authenticationMethod": "Authentication method",
   "initial.handle.email": "Email address",
   "initial.handle.phone": "Phone number",
-  "initial.handle.phone.email": "Type your email",
+  "initial.handle.email.placeholder": "Type your email",
   "initial.handle.phone.placeholder": "Type your phone number",
   "initial.submit": "Continue",
   "initial.divider": "or",
--- a/packages/react/src/components/text/constants.ts
+++ b/packages/react/src/components/text/constants.ts
@@ -196,4 +196,9 @@ export const TEXT = {
     "Cookies that track your online behaviour, such as clicks, preferences, device specifications, location, and search history. This data helps in targeted advertising and gathering website analytics.",
   "gdpr.dialog.error.title": "Oops!",
   "gdpr.dialog.error.subtitle": "Looks like something went wrong...",
+  // deprecated keys
+  /**
+   * @deprecated Use 'initial.handle.email.placeholder' instead
+   */
+  "initial.handle.phone.email": "",
 };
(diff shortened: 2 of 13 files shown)


packages/react-primitives/src/components/tabs/index.tsx
```diff
@@ -21,6 +21,7 @@ export const Tabs: React.FC<Props> = ({ className, tabs, defaultValue }) => {
 
   return (
     <RadixTabs.Root
+      data-testid="sid-handle-type-tabs"
```
ikovic, line 24: This is a primitive component, so the test ID should be passed in as a prop.

packages/react/src/domain/handles.ts
```diff
@@ -29,6 +30,10 @@ const SSO_FACTORS = ["oidc", "saml"];
 function getPossibleHandleTypes(factor: Factor): Set<HandleType> {
   const handleTypes = new Set<HandleType>();
 
+  if (isFactorWithAllowedHandleTypes(factor)) {
```
ikovic, line 33: This order of checks allows a weird situation where you specify the extra option to use a specific handle type before checking if the given factor actually requires that handle type.

To be on the safe side, I would turn this around so that you still check if the factor actually requires the handle type specified in the `allowedHandleTypes` thing and only add it in these cases:
- there is no `allowedHandleTypes` prop present
- `allowedHandleTypes` is present and the given type is included

ikovic:
@kaspermroz two general thoughts:
- we should document this option with a follow up PR
- I am not sure if the name `allowedHandleTypes` is the best choice

More on the second point above - without looking at the docs, would you assume that an undefined or an empty array would allow all handle types or none? The property is optional, but I feel like it might be confusing for folks who see it for the first time.

Also thinking about it, I'm pretty sure this would be a frequent requirement as a lot of companies would work with email exclusively.

Maybe there's no need to change anything though, I just want to make sure the API is self describing. How does the TypeScript feedback look like if you try specifying this for an unsupported factor?

packages/react/src/domain/handles.ts
```diff
@@ -29,6 +30,10 @@ const SSO_FACTORS = ["oidc", "saml"];
 function getPossibleHandleTypes(factor: Factor): Set<HandleType> {
```
ikovic, line 30: Should we change the signature of this function? Looks like it can receive types other than `Factor`?

packages/react/src/domain/types.ts
```diff
@@ -62,6 +62,17 @@ export type FactorEmailLink = Extract<Factor, { method: "email_link" }>;
 
 export type FactorSmsLink = Extract<Factor, { method: "sms_link" }>;
 
+/**
+ * Utility type to specify allowed handle types in case given factor supports more than one.
+ */
+export type FactorWithAllowedHandleTypes = Factor & {
+  /**
+   * Limits the allowed handle types for given factor method to the ones specified in the array.
+   * When this option is not specified, all factors are allowed.
```
ikovic, line 71: `all factors are allowed` => `all supported handle types are allowed`

packages/react/src/domain/handles.ts
```diff
@@ -26,15 +28,28 @@ const FACTORS_WITH_EMAIL = [
...
+      !isFactorWithAllowedHandleTypes(factor) ||
+      (isFactorWithAllowedHandleTypes(factor) &&
+        factor.allowedHandleTypes?.includes("email_address"))
+    )
+      handleTypes.add("email_address");
   }
 
   if (FACTORS_WITH_PHONE.includes(factor.method)) {
-    handleTypes.add("phone_number");
+    if (
```
ikovic, line 46: There's a bit of duplication now, how do you feel about extracting the new check in a separate function?

`function isHandleTypeAllowed(factor: Factor | FactorWithAllowedHandleTypes, handleType: HandleType): boolean` could help make this more readable

packages/react/src/domain/types.ts
```diff
@@ -110,3 +125,11 @@ export type ValidationError = {
 export type Validator<T = unknown> = (value: T) => ValidationError | undefined;
 
 export type MaybeArray<T> = T | T[];
+
+export function isFactorWithAllowedHandleTypes(
+  factor: Factor
+): factor is FactorWithAllowedHandleTypes {
+  return Object.hasOwn(factor, "allowedHandleTypes");
```
ikovic, line 132: Should we assert this is an array with the correct items? This value comes from the developer, so it may be incorrect.

packages/react/src/components/text/constants.ts
```diff
@@ -11,7 +11,7 @@ export const TEXT = {
   "initial.authenticationMethod": "Authentication method",
   "initial.handle.email": "Email address",
   "initial.handle.phone": "Phone number",
-  "initial.handle.phone.email": "Type your email",
+  "initial.handle.email.placeholder": "Type your email",
```
jake-slashid, line 14: Is this a breaking change? 🤔

jake-slashid (approved):

jake-slashid (changes requested):
Lets discuss breaking change remediation

ikovic (approved):

jake-slashid (approved):

PR #219, opened 2024-02-14
[SID-1421][SID-1529] Improve OTP error & resend handling

kaspermroz:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1421)
[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1529)

Changes:
 - Create <Delayed> component in `react-primitives`
 - Clear input and show error message when incorrect OTP code is submitted
 - Transition to `retry` state on retry, add corresponding UI

Testing:
 - Authenticate with OTP via email
 - Enter incorrect code
 - Input should be cleared and error message displayed
 - Resend code
 - You should see temporary retry UI
 - Once you get the code, you should see the fresh empty input state

## Checklist

### DIFF
--- a/packages/react-primitives/src/components/delayed/index.tsx
+++ b/packages/react-primitives/src/components/delayed/index.tsx
@@ -1,0 +1,21 @@
+import { useState, useEffect, type ReactNode } from "react";
+
+type Props = {
+  delayMs: number;
+  children: ReactNode;
+  fallback?: ReactNode;
+};
+
+export function Delayed({ delayMs, children, fallback }: Props) {
+  const [render, setRender] = useState(false);
+
+  useEffect(() => {
+    const timeout = setTimeout(() => {
+      setRender(true);
+    }, delayMs);
+
+    return () => clearTimeout(timeout);
+  }, [delayMs]);
+
+  return <>{render ? children : fallback ?? null}</>;
+}
--- a/packages/react-primitives/src/components/delayed/index.tsx
+++ b/packages/react-primitives/src/components/delayed/index.tsx
@@ -1,12 +1,19 @@
 import { useState, useEffect, type ReactNode } from "react";
 
 type Props = {
+  /** Period of time, after which the component will render its children */
   delayMs: number;
   children: ReactNode;
+  /** Optional fallback component rendered initially, replaced by children after the delay */
   fallback?: ReactNode;
+  /** Optional CSS class name for the wrapper */
+  className?: string;
 };
 
-export function Delayed({ delayMs, children, fallback }: Props) {
+/**
+ * Utility component used to render its children after specified period of time
+ */
+export function Delayed({ delayMs, children, fallback, className }: Props) {
   const [render, setRender] = useState(false);
 
   useEffect(() => {
@@ -17,5 +24,7 @@ export function Delayed({ delayMs, children, fallback }: Props) {
     return () => clearTimeout(timeout);
   }, [delayMs]);
 
-  return <>{render ? children : fallback ?? null}</>;
+  return (
+    <div className={className}>{render ? children : fallback ?? null}</div>
+  );
 }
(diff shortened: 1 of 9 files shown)


packages/react-primitives/src/components/delayed/index.tsx
```diff
@@ -0,0 +1,21 @@
+import { useState, useEffect, type ReactNode } from "react";
+
+type Props = {
+  delayMs: number;
+  children: ReactNode;
+  fallback?: ReactNode;
+};
+
+export function Delayed({ delayMs, children, fallback }: Props) {
```
ikovic, line 9: Some TS docs would be great + maybe have an optional `className` prop? I can imagine a use case where we wanted to animate the transition

packages/react/src/components/form/authenticating/messages.ts
```diff
@@ -4,8 +4,11 @@ import { TextConfigKey } from "../../text/constants";
 // TODO add case for password
 export function getAuthenticatingMessage(
   factor: Factor,
-  isSubmitting = false
+  formState: "initial" | "input" | "submitting" | "retry" = "initial"
 ): { title: TextConfigKey; message: TextConfigKey } {
+  const isSubmitting = formState === "submitting";
```
ikovic, line 9: Do these variables add value? Looking at the form state seems quite explicit, as the whole intent was to enable easy conditional rendering.

packages/react/src/components/form/authenticating/otp.tsx
```diff
@@ -42,17 +42,21 @@ const FactorIcon = ({ factor }: { factor: Factor }) => {
...
+    values,
+    registerField,
+    registerSubmit,
+    setError,
+    hasError,
+    clearError,
+  } = useForm();
   const [formState, setFormState] = useState<
-    "initial" | "input" | "submitting"
+    "initial" | "input" | "submitting" | "retry"
```
ikovic, line 54: We might not need this extra state here - if you inspect the current flow state, you can check if we have retried already. I believe there is a retry counter in the context.

jake-slashid (approved):

packages/react/src/components/form/authenticating/otp.tsx
```diff
@@ -79,11 +95,19 @@ export const OTPState = ({ flowState }: Props) => {
...
+        clearError("otp");
+      }
       onChange(event as never);
     },
-    [registerField, text]
+    [clearError, hasError, registerField, text, values]
   );
 
+  const handleRetry = () => {
+    setFormState("retry");
```
ikovic, line 107: If we [get rid of the local retry state](https://github.com/slashid/javascript/pull/219/files#r1489300244), this wouldn't be needed anymore.

packages/react/src/components/text/constants.ts
```diff
@@ -90,6 +92,8 @@ export const TEXT = {
   "authenticating.title.oidc": "Sign in with ",
   "authenticating.otpInput": "OTP",
   "authenticating.otpInput.submit": "Submit",
+  "authenticating.itpInput.submit.error":
```
ikovic, line 95: Typo here - `itpInput`

packages/react/src/components/text/constants.ts
```diff
@@ -90,6 +92,8 @@ export const TEXT = {
...
+    "Incorrect OTP code submitted! Insert valid code.",
```
ikovic, line 96: `Please enter a valid code` should be sufficient

packages/react/src/components/form/authenticating/messages.ts
```diff
@@ -36,6 +44,12 @@ export function getAuthenticatingMessage(
       };
     }
     case "otp_via_email":
```
ikovic, line 46: What about `otp_via_sms`?

ikovic (approved):

PR #221, opened 2024-02-28
React: use alternative to Object.hasOwn() to improve browser support

jake-slashid:
This should really happen at the implementation site, but I think the bundle size trade-off is worth it for more wide reaching stability at this time.

I ran the formatter so there are some unrelated changes in this PR.


paulo-raca (approved):

ikovic (changes requested):
@jake-slashid seems like there are unrelated changes in the PR, can you check please?

My suggestion is to refactor the offending code to this:
```
if (!("allowedHandleTypes" in factor)) {
    return false;
  }
```

IMO this comes with less baggage

jake-slashid:
@ikovic 
> I ran the formatter so there are some unrelated changes in this PR.

ikovic (approved):

PR #223, opened 2024-03-01
React: conditionally render passkey method based on platform support

jake-slashid:
Adding passkey support detection to the auth method selector in `<Form>`

This PR is just barely working, full of debugs and in need of some TLC.

Coming soon!

### DIFF
--- a/packages/react/src/components/form/initial/controls.tsx
+++ b/packages/react/src/components/form/initial/controls.tsx
@@ -24,6 +24,7 @@ import { useForm } from "../../../hooks/use-form";
 import { TextConfig, TextConfigKey } from "../../text/constants";
 import { ErrorMessage } from "../error-message";
 import { isValidEmail, isValidPhoneNumber } from "../authenticating/validation";
+import { passkeysSupported } from '../../../passkey'
 
 import * as styles from "./initial.css";
 import { useInternalFormContext } from "../internal-context";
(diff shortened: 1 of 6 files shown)


kaspermroz (approved):

packages/react/src/passkey.ts
```diff
@@ -0,0 +1,3 @@
+import { Utils } from "@slashid/slashid";
+
+export const runtimeSupportsPasskeys = Utils.runtimeSupportsBestPasskeyUX();
```
ikovic, line 3: This might cause trouble if executed server side. Can we make sure this either runs client side only or gets parametrised so the user agent string is passed in? Server side we can read it from the HTTP request.

packages/react/src/components/form/initial/controls.tsx
```diff
@@ -172,16 +173,33 @@ const FormInput = ({ children }: FormInputProps) => {
...
+    [factors, showPasskeys]
   );
 
   const handleTypes = useMemo(() => {
     return getHandleTypes(factors);
   }, [factors]);
 
+  const passkeySupportDetectionLoading = showPasskeys === null;
+  if (passkeySupportDetectionLoading) {
+    return null;
```
ikovic, line 200: Ideally this would be a skeleton or some sort of placeholder, so we can avoid layout shifts.

jake-slashid: Agreed

PR #224, opened 2024-03-06
[SID-1400] Add `environment` prop to `<SlashIDProvider>`

kaspermroz:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1400)

The Core SDK have changed the environment config API some time ago, we want to update React SDK to enable the same configuration.

## Checklist

### DIFF
--- a/packages/react/src/context/slash-id-context.tsx
+++ b/packages/react/src/context/slash-id-context.tsx
@@ -8,7 +8,12 @@ import {
   useState,
 } from "react";
 
-import { PersonHandleType, SlashID, User } from "@slashid/slashid";
+import {
+  PersonHandleType,
+  SlashID,
+  SlashIDEnvironment,
+  User,
+} from "@slashid/slashid";
 import {
   ThemeProps,
   ThemeRoot,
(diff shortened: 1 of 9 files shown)


packages/react/src/context/slash-id-context.tsx
```diff
@@ -234,6 +259,7 @@ export const SlashIDProvider = ({
     if (state === "initial") {
       const slashId = new SlashID({
         oid,
+        ...(environment && { environment }),
```
kaspermroz, line 262: I didn't add any sort of validation since the Core SDK handles that already, I think marking the other properties as `@deprecated` is enough - let me know WDYT @ikovic @jake-slashid

ikovic: @kaspermroz I think the validation part is OK, core SDK will do that. What happens for customers that upgrade to this version without changing to the new format? Will TS complain a lot? Maybe a union of props would be a better idea? To make sure `environment` cannot be used in the combination with old props.

packages/react/src/context/slash-id-context.tsx
```diff
@@ -234,6 +259,7 @@ export const SlashIDProvider = ({
...
         ...(baseApiUrl && { baseURL: baseApiUrl }),
```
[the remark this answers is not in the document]
kaspermroz: I know but this throws an error on the Core SDK level and this is fine to me. 
TypeScript marks the urls as deprecated so people will know that they should use the `environment` instead. Would you rather this to filter out the urls if environment is specified instead?

packages/react/src/context/slash-id-context.tsx
kaspermroz: We can do that, this will give you a nice TypeScript error

kaspermroz: Tried that, the problem is that you lose the nice DX with deprecation warning and JS docs. I'd rather leave that as is and have the deprecation warning + runtime error if you pass in the two options

packages/react/src/context/slash-id-context.tsx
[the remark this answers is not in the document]
ikovic: All good, thanks! I was pretty sure I saw the `baseApiUrl` being passed in unconditionally, this is good!

ikovic (approved):

PR #225, opened 2024-03-06
React: add Customisation namespace

jake-slashid:
Adds new `Customisation` namespace, with a handful of helpers for the hosted login page customisation.

New utils:
- `getGoogleFontImports`
- `sanitiseCssVariableCustomisationConfig`
- `pxValidator`
- `hexValidator`
- `exactValidator`
- `fontFamilyValidator`


packages/react/src/utils/css-sanitisation.ts
```diff
@@ -0,0 +1,76 @@
+export const pixelValueRegExp = /([0-9]+)px/;
+export const hexValueRegExp = /#([\da-fA-F]{2})([\da-fA-F]{2})([\da-fA-F]{2})/;
+export const fontFamilyRegExp = /(?:['"]?([\\\w\d\- ]+)['"]?(?:,\s*)?)/g;
+export const displayValues = ["flex", "none"];
+export const googleFonts = new Set(["Open Sans"])
```
ikovic, line 5: Pretty sure that `Inter` is also a google font

packages/react/src/dev.css
```diff
@@ -30,9 +30,9 @@ body {
 }
 
 .layout {
-  display: grid;
+  /* display: grid;
   grid-template-columns: 1fr 1fr;
-  gap: 1rem;
+  gap: 1rem; */
```
kaspermroz, line 35: Please remove commented out code

packages/react/src/dev.tsx
```diff
@@ -163,38 +163,90 @@ const ConfiguredDynamicFlow = () => {
...
   return (
     <ConfigurationProvider
+      logo="https://membersapp.checkatrade.com/static/media/logo.e930e0049182a5b0e5bf.png"
       factors={[
-        { method: "webauthn" },
-        { method: "email_link" },
-        { method: "otp_via_email" },
-        { method: "otp_via_sms" },
-        { method: "password" },
+        // { method: "webauthn" },
```
kaspermroz, line 168: Bunch of mess in this file, I'd rather not commit these changes

kaspermroz (commented):
In general I don't think we should create `otherPublicVariables` as a interface, it should be unified and with as little vars as possible to do the job

packages/react/src/dev.css
ikovic: Done already in the previous commit

packages/react/src/utils/css-sanitisation.ts
ikovic: Fixed!

packages/react/src/dev.tsx
ikovic: None of these will be committed, I'll clean this up after I'm done testing

kaspermroz (approved):
Great job guys! 🎖️

PR #226, opened 2024-03-14
React: add wrapper for image logo

jake-slashid:
Add a wrapper around the `` element, allowing flexbox alignment without hurting the images natural size

There is a discrepancy between the markup for img vs /id (svg) logo:
```html
<div className="[logo-classes]">
  <Logo>
</div>
```

vs

```html
<img className="[logo-classes]" />
```

In the case of `img`, it's not possible to use flex alignment CSS because there is no flex container around the logo itself.

### DIFF
--- a/packages/react/src/components/form/initial/logo.tsx
+++ b/packages/react/src/components/form/initial/logo.tsx
@@ -12,11 +12,15 @@ export type Props = {
 export const Logo: React.FC<Props> = ({ logo }) => {
   if (typeof logo === "string" && logo) {
     return (
-      <img
-        className={clsx("sid-logo", "sid-logo--image", styles.logo)}
-        src={logo}
-        alt="Company logo"
-      />
+      <div className={clsx("sid-logo", "sid-logo--image", styles.logo)}>
+        <div>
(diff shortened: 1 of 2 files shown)


paulo-raca (approved):

packages/react/src/components/form/initial/logo.tsx
```diff
@@ -12,11 +12,15 @@ export type Props = {
 export const Logo: React.FC<Props> = ({ logo }) => {
   if (typeof logo === "string" && logo) {
     return (
-      <img
-        className={clsx("sid-logo", "sid-logo--image", styles.logo)}
-        src={logo}
-        alt="Company logo"
-      />
+      <div className={clsx("sid-logo", "sid-logo--image", styles.logo)}>
+        <div>
```
paulo-raca, line 16: Do we need 2 nested divs?

paulo-raca: No idea what a flex parent-child thing do, but you seem to know :+1:

ikovic (approved):

PR #233, opened 2024-04-05
[SID-1647] Run e2e tests in Docker

ikovic:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1647)

Run the e2e tests in Docker. Use the official Playwright image. An email account was set up on `mailinator` so we can test email as part of the flow.

This only covers the `react-nextjs` app, but it also sets up the infrastructure needed to easily test other projects as well.

## Next steps

- add more apps to the matrix (right now it's just `react-nextjs` on `chromium`)
- improve test coverage
- decide when to run these tests
- when tests involve mutating persistent state, create a new suborg before the test

## Checklist


Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM node:20-slim
+FROM mcr.microsoft.com/playwright:v1.42.1-jammy
```
ikovic, line 2: `node` from the first FROM will not be used

ikovic: Resolved

packages/tests/e2e/auth.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+
+test.describe("Authentication", () => {
+  test("Log in with email link", async ({ page, context }) => {
+    const inbox = `e2e-${generateRandomNumber()}`;
+
+    await page.goto("http://localhost:3000/form");
+
+    await page
+      .locator("#sid-input-email_address")
+      .fill(`${inbox}@team336427.testinator.com`);
```
paulo-raca, line 21:
```suggestion
      .fill(`${inbox}@${TEAM_DOMAIN}`);
```

packages/tests/src/email.ts
```diff
@@ -0,0 +1,108 @@
...
+export function getTheJumpPageURL(email: string) {
+  const $ = load(email);
+  const jumpPageLink = $("a:contains('Confirm')").attr("href");
+  return jumpPageLink;
+}
+
+export async function fetchLatestEmailWithRetry(
+  input: FetchLatestEmailInput
+): Promise<Message | null> {
+  const maxRetries = 5;
```
paulo-raca, line 89: Emails are slow and asynchronous. 5 retries doesn't seem to be enough.

My guess is that it should retry for at least 30s, possibly more.

packages/tests/src/email.ts
```diff
@@ -0,0 +1,108 @@
...
+    console.log("Error fetching email", e);
+    return null;
+  }
+}
+
+export type GetMessageInput = {
+  messageId: string;
+};
+
+export async function getMessage({ messageId }: GetMessageInput) {
```
paulo-raca, line 48: nit: maybe merge `fetchLatestEmail` and getMessage into a single function?

packages/tests/playwright.config.ts
```diff
@@ -44,15 +47,15 @@ export default defineConfig({
       use: { ...devices["Desktop Chrome"] },
     },
 
-    {
+    /* {
```
paulo-raca, line 50: Double checking: Do we really want those disabled?

packages/tests/src/email.ts
```diff
@@ -0,0 +1,108 @@
...
+  try {
+    const mailinatorClient: MailinatorClient = new MailinatorClient(
+      "4c39b2cfe3c746a09926ca6e0a17c779"
```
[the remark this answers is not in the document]
ikovic: Yep, I added the ENV var just to prevent this but forgot to remove it :) It's easy to rotate so I'll do it immediately

packages/tests/playwright.config.ts
ikovic: I disabled them at least until code review is done in order to iterate quickly. We still have to discuss the entire testing matrix (browser, build tool, UI rendering framework) - these multiply the number of runs so I wanted to check with the team what they think about the trade offs.

packages/tests/src/email.ts
ikovic: Agreed, I did not tweak these things - just wanted to get a running baseline for now.

apps/react-nextjs/.env
```diff
@@ -1 +1,4 @@
-NEXT_PUBLIC_ORG_ID=
\ No newline at end of file
+# internal_testing_2
+NEXT_PUBLIC_ORG_ID=00000000-0000-0000-0200-000000000000
```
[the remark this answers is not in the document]
ikovic: We talked to @balenio about this and agreed we should use the internal testing orgs since we already have a cleanup script that removes old suborgs from the testing environment.

Once we start writing tests that mutate persistent state (eg writing to attributes), each test run will create a separate suborg in order to ensure clean state.

packages/tests/e2e/auth.spec.ts
```diff
@@ -0,0 +1,62 @@
+import { test, expect } from "@playwright/test";
+import {
+  fetchLatestEmailWithRetry,
+  getMessage,
+  getTextContent,
+  getTheJumpPageURL,
+} from "../src/email";
+
+const generateRandomNumber = (): number => {
+  return Math.floor(Math.random() * 9000) + 1000;
```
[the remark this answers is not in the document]
ikovic: Makes sense, the service I am using clears inboxes but it is entirely possible to get two same numbers in a row.

packages/tests/e2e/auth.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+};
+
+test.describe("Authentication", () => {
+  test("Log in with email link", async ({ page, context }) => {
+    const inbox = `e2e-${generateRandomNumber()}`;
```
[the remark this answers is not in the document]
ikovic: Will be better definitely, I'll take another pass over that module to make it easier to use. I was building it step by step so it evolved into something a bit messy.

paulo-raca (approved):

roblaszczak (approved):

PR #234, opened 2024-04-05
[SID-1591] react: add support for anon users

jake-slashid:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1591-anon-users)

Adds support for anonymous users
- Add `anonymousUsersEnabled` to `SlashIDProvider`
- During initialisation:
    1. If token stored: create anon user or user based on token
    2. If no token: try to login with directid
    3. If no direct id and no token: create anon user (if enabled)
- `logIn()` considers if user is a `AnonymousUser` or logged out
  - If anon user, upgrade to full user with `AnonymousUser.id()`
  - If logged out, login with `sid.id()`
- `mfa()` considers if user is regular user or anon user
- During token validation, if token is `person_type === 'anonymous'` and anon users disabled: token validation fails


packages/react/src/utils/anonymous.ts
```diff
@@ -0,0 +1,5 @@
+import { BaseUser, AnonymousUser } from "@slashid/slashid";
+
+export const userIsAnonymous = (user: BaseUser): user is AnonymousUser => {
```
ikovic, line 3: Wondering if this is needed, what is the added value on top of working with the `user` directly and checking `user.anonymous`? It's a documented public property

packages/react/src/context/slash-id-context.tsx
```diff
@@ -52,13 +54,14 @@ export interface SlashIDProviderProps {
    */
   sdkUrl?: string;
   analyticsEnabled?: boolean;
+  anonymousUsersEnabled?: boolean;
```
ikovic, line 57: It would be useful to annotate this property with a TSDoc style comment, this will pop up in the editor - could save someone a visit to the docs.

packages/react/src/context/slash-id-context.tsx
```diff
@@ -201,25 +205,42 @@ export const SlashIDProvider = ({
...
         throw e;
       }
     },
-    [oid, state, storeUser, logOut]
+    [state, storeUser, user, oid, logOut]
   );
 
   const mfa = useCallback<MFA>(
     async ({ handle, factor }) => {
+      if (user && userIsAnonymous(user)) {
```
ikovic, line 237: We need the same check on the `recover` method as it cannot be called with an anonymous user.

packages/react/src/context/slash-id-context.tsx
```diff
@@ -321,12 +350,22 @@ export const SlashIDProvider = ({
...
+          storeUser(user);
+        }
       } else {
         const isDone = await loginDirectIdIfPresent();
 
-        if (!isDone) {
+        if (!isDone && anonymousUsersEnabled) {
+          const user = await sid.createAnonymousUser();
+          storeUser(user);
+        } else if (!isDone) {
```
ikovic, line 368: @jake-slashid if I'm reading this right, enabling anonymous users will prevent the SDK from reading tokens from storage.

We should only create an anonymous user if we have already looked into the storage and found it empty.

This function was a bit messy before adding anonymous users, I was wondering if we could refactor it in a way that makes it clear what the priority is? I've had success in previous projects following the chain of responsibility pattern:

```
const chain = [
 useTokenFromProps,
 useDirectID,
 useStoredToken,
 useAnonymousUser
]

// execute chain in sequence
// stop execution once the first function returns a truthy value
// each function can either return a Promise that resolves to a User or resolve with `false` or similar
```
This way the priority and the sequence of steps is made clear. What do you think?

jake-slashid: I agree that the reasonability of this block is very poor, but its intended that the anon user creation is last.

jake-slashid: I’ll refactor this function

packages/react/src/utils/anonymous.ts
jake-slashid: It’s for type narrowing, so we don’t need to cast.

`User | AnonymousUser | undefined` -> `AnonymousUser`

Anon users have a unique method: `id()`. Calling this safely was the primary goal, anywhere else I used it was just for consistency.

jake-slashid:
I'm leaning towards `Groups` supporting anon user groups.

My thought process:
- `Groups` has a single responsibility: assert that a user has a group.
- Previously it did coincidentally also assert that a user was logged in, so you could in theory use it for that too.
- `LoggedIn` has the single responsibility: assert that a user is authenticated/not anonymous.
- Although verbose, I think it's better for the consumer to compose these two logic atoms to get their desired outcome.

```tsx
<LoggedIn>
  <Groups belongsTo="admin">
    admin & user
  </Groups>
</LoggedIn>

<LoggedOut>
  <Groups belongsTo="admin">
    admin & anon user
  </Groups>
</LoggedOut>

<Groups belongsTo="admin">
  admin & (user | anon user)
</Groups>
```

@ikovic

packages/react/src/context/slash-id-context.tsx
jake-slashid: done

packages/react/src/context/slash-id-context.tsx
jake-slashid: done

There is a new utility: `sequence()`

jake-slashid:
A lot of unit tests are failing, looking into it

packages/react/src/utils/anonymous.ts
ikovic: @jake-slashid is this correct though? A `BaseUser` that is anonymous could have `user.anonymous` set to `true`, but not be an instance of `AnonymousUser` 🤔

packages/react/src/components/logged-in/index.tsx
```diff
@@ -27,7 +30,7 @@ export const LoggedIn: React.FC<Props> = ({ children, withFactorMethods }) => {
 
   const shouldRender = useMemo(() => {
     // user must be logged in
-    if (!isAuthenticated || !user) {
+    if (!isAuthenticated || !user || user.anonymous) {
```
ikovic, line 33: Makes sense to me! This might be a good candidate for a unit test

packages/react/src/hooks/use-slash-id.ts
```diff
@@ -13,7 +13,7 @@ export function useSlashID(): UseSlashID {
     [contextValue.sdkState]
   );
   const isAuthenticated = useMemo(
-    () => contextValue.user !== undefined,
+    () => contextValue.user !== undefined && !contextValue.user.anonymous,
```
jake-slashid, line 16: This is a bit of a controversial change IMO

I decided this made sense because `isAuthenticated` previously meant "user instanceof User", but before this change the meaning was `user instanceof User || user instanceof AnonymousUser`

It took me a while to decide on this because `isAuthenticated` is kind of a vague variable name, technically speaking anon users are authenticated and can do things in the API, but they're not authenticated in terms of customer application logic.

packages/react/src/utils/anonymous.ts
jake-slashid: Yeah good point!

packages/react/src/components/logged-in/index.tsx
jake-slashid: I've actually changed my mind on this, it conflicts with the assumption made in `Groups`:
https://github.com/slashid/javascript/pull/234#issuecomment-2044794656

jake-slashid:
```suggestion
    if (!isAuthenticated || !user) {
```

ikovic: This is a tough question and I was asking @balenio and @paulo-raca for help on determining this. We ended up on this being the definition of an authenticated user:

```
token is valid
person_type != "anonymous" (To defend from old token that had a missing field)
```

[Relevant discussion here](https://slashidworkspace.slack.com/archives/C048HJ7D5FD/p1710422233999519)

jake-slashid: Looks like we are aligned!

https://github.com/slashid/javascript/pull/234/files#r1559619822

ikovic:
## Testing notes

Using the following versions to run the test in a new Vite React app:

```
"@slashid/react": "1.23.0-anonymous.1",
"@slashid/slashid": "3.23.0-next.1",
```

### Analytics
A `PersonIdentified_v1` event is emitted after obtaining an anonymous user. Is this user really identified or should we emit a different kind of event? Nothing in the event tells that the person is anonymous.

ikovic (approved):

PR #235, opened 2024-04-09
[SID-1650]React SDK: Implement an e2e test to verify the OTP via email flow

ikovic:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1650)

Added an e2e test that verifies the happy path of the `otp_via_email` flow. Added extra - implemented a page object model for the `/form` page in the demo app. If this looks good I'll refactor the existing test with this approach.

## Checklist

### DIFF
--- a/packages/tests/package.json
+++ b/packages/tests/package.json
@@ -5,7 +5,7 @@
   "description": "Tests setup for /id projects",
   "main": "index.js",
   "scripts": {
-    "test:e2e": "playwright test"
+    "test:e2e": "playwright test --debug"
   },
   "author": "",
   "license": "ISC",
@@ -18,6 +18,7 @@
     "playwright.config.ts"
   ],
   "dependencies": {
+    "@slashid/slashid": "^3.22.0",
     "cheerio": "1.0.0-rc.12",
     "dotenv": "^16.4.5",
     "mailinator-client": "^1.0.4",
--- a/packages/tests/src/pom/form.ts
+++ b/packages/tests/src/pom/form.ts
@@ -1,0 +1,56 @@
+import { Locator, type Page } from "@playwright/test";
+import type { FactorMethod } from "@slashid/slashid";
+
+type FactorMethodLabel = "OTP via email" | "Email link";
+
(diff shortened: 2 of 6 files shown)


packages/tests/package.json
```diff
@@ -5,7 +5,7 @@
   "description": "Tests setup for /id projects",
   "main": "index.js",
   "scripts": {
-    "test:e2e": "playwright test"
+    "test:e2e": "playwright test --debug"
```
kaspermroz: I guess we shouldn't commit this?

kaspermroz (approved):

packages/tests/package.json
ikovic: pretty sure you caught an older version, I removed it 😓

packages/tests/package.json
```diff
@@ -18,6 +18,7 @@
     "playwright.config.ts"
   ],
   "dependencies": {
+    "@slashid/slashid": "^3.22.0",
```
ikovic, line 21: Added this so we can use the types from here, it's a better DX when writing tests

packages/tests/src/pom/form.ts
```diff
@@ -0,0 +1,56 @@
...
+  // @ts-expect-error add missing labels when needed
+  const methodToLabel: Record<FactorMethod, FactorMethodLabel> = {
+    otp_via_email: "OTP via email",
+    email_link: "Email link",
+  };
+
+  return methodToLabel[method];
+}
+
+export class FormPage {
```
ikovic, line 16: Page object model for the `/form` page. High level idea:
- test scenarios should not depend on locators as they are low level
- instead test scenarios depend on page object models
- a page object model hides away the locators and exposes a high level API
- test scenario can then be written using a more natural language based on the page object model

ikovic: At the same time this makes the test scenarios more readable and also easier to maintain, given that multiple scenarios can target the same page. This way they will reuse the same selectors instead of duplicating them.

jake-slashid (approved):

PR #236, opened 2024-04-12
Introduce state charts to the login flow

kaspermroz:
## Description

This PR introduces a nested state to the Authenticating flow state. 

### Context

The current login flow implementation looks as follows:
 - we call the id() method
 - transition to authenticating state
 - setup event listeners on the auth state
 - wait for SDK events and proceed

For TOTP, as soon as we get the response from /id method, the event is emitted almost instantly, so the order is:
 - we call id()
 - event is fired
 - we transition to authenticating state and setup listeners for the event that's already emitted, so we're never exit a 
 - particular UI state

It makes sense to move the UI state orchestration bit with event listeners to the flow itself and let the UI be driven by that state machine - we end up with a state chart, which is basically a nested state machine - flow is a state machine itself, now the authenticating state is going to be chunked into couple states, so it becomes a state machine as well.

## Checklist


packages/react/src/components/form/ui-state-machine.ts
```diff
@@ -0,0 +1,182 @@
...
+  }
+}
+
+class OTPUIStateMachine extends UIStateMachine<OTPStatus> {
+  constructor(opts: UIStateMachineOpts) {
+    super(opts);
+    this.state = this.prepareNextState("initial");
+  }
+
+  private prepareNextState(status: OTPStatus): State<OTPStatus> {
```
ikovic, line 90: This naming breaks my brain 😅

What does this method do? Seems a bit confusing since it seems to call itself as part of `this.transition` calls.

Usually the API designed around state machines would have a `transition` function that receives an event that is supposed to transition the machine to the next state. This is evaluated against the current state to determine if the transition is possible.

kaspermroz: Some context here - I initially split the `transition()` into two methods:
 - `prepareNextState(status: "intial" |  "input" | ... )` is responsible for creating the appropriate `State` object based on the status, it has the `switch` instruction inside
 - `transition(state: State)` takes in the `State` object, swaps the state and fires the `entry` and `exit` events
 
This is probably not the best way to do this though, I'm open for refactoring this as I'm doing bit of experimenting here

packages/react/src/components/form/authenticating/otp.tsx
```diff
@@ -46,26 +41,24 @@ export const OTPState = ({ flowState }: Props) => {
...
-    isSubmitting: formState === "submitting",
+    isSubmitting: flowState.hasUIState("submitting"),
     hasRetried,
   });
 
   const handleSubmit: FormEventHandler<HTMLFormElement> = useCallback(
     (e) => {
       e.preventDefault();
+      const uiState = flowState.getUiState();
+      if (!isInputState(uiState)) return;
```
ikovic, line 57: This is supposed to be done in the state machine, UI should not be worried about this

packages/react/src/components/form/flow.ts
```diff
@@ -27,6 +32,8 @@ export interface AuthenticatingState {
   cancel: Cancel;
   recover: () => void;
   entry: () => void;
+  hasUIState: (state: AuthenticatingUIStatus) => boolean;
```
ikovic, line 35: Using this naming convention is limiting us a bit, since the nested state does not have to be a UI state (eg it could be an ephemeral state that does some side effect).

Maybe doing something similar to [XState.matches](https://xstate.js.org/docs/guides/states.html#state-methods-and-properties) could help here? So you can pass the parent state name or the child state name, it should return the correct result anyway?

packages/react/src/components/form/flow.ts
```diff
@@ -27,6 +32,8 @@ export interface AuthenticatingState {
...
+  getUiState: () => State<AuthenticatingUIStatus>;
```
ikovic, line 36: On a similar note, maybe consider a name that does not imply this must be a UI state? `getNestedState` or `getChildState` maybe?

packages/react/src/components/form/authenticating/otp.tsx
kaspermroz: Yes but this function kind of doubles as a type guard, so TypeScript doesn't complain about using `submit()` method. How should I adjust it then, by type casting?
```ts
(flowState as LoginState).submit(...)
```

ikovic: I guess it's better to leave it as it is then!

packages/react/src/components/form/ui-state-machine.ts
```diff
@@ -0,0 +1,343 @@
...
+
+      case "sid_ui.passwordRecoveryRequested":
+        this.transition({
+          status: "recoverPassword",
+          entry: async () => {
+            try {
+              await this.recover();
+              this.send({ type: "sid_ui.passwordRecovered" });
+            } catch (e) {
+              // ignored
```
ikovic, line 250: Is this error caught somewhere else? In the existing flow, failing to recover a password should send you to the error state:
```
catch (error) {
      send({ type: "sid_login.error", error: ensureError(error) });
    }
```

packages/react/src/components/form/ui-state-machine.ts
```diff
@@ -0,0 +1,343 @@
...
+      status: "initial",
+      entry: () => {
+        this.logInFn(this.context.config, this.context.options)
+          .then((user) => {
+            if (user) {
+              this.send({ type: "sid_login.success", user });
+            } else {
+              this.send({
+                type: "sid_login.error",
+                error: new Error("User not returned from /id"),
```
ikovic, line 280: Since we're at it, it might make sense to introduce the `SlashIDError` here instead of the basic `Error` type?

packages/react/src/components/form/authenticating/otp.tsx
```diff
@@ -104,7 +97,7 @@ export const OTPState = ({ flowState }: Props) => {
   const handleRetry = () => {
     flowState.retry();
     clearError("otp");
-    setFormState("submitting");
+    // setFormState("submitting");
```
ikovic, line 100: Ready to be removed

packages/react/src/components/form/flow.ts
```diff
@@ -42,53 +49,16 @@ export interface ErrorState {
...
   // a function that will be called when the state is entered
   entry?: () => void;
 };
 
 export type FlowState = FlowActions &
   (InitialState | AuthenticatingState | SuccessState | ErrorState);
 
-type Observer = (state: FlowState, event: Event) => void;
-type Send = (e: Event) => void;
+export type Observer = (state: FlowState, event: Event) => void;
```
ikovic, line 60: Should this be moved to the dedicated types module?

ikovic:
@kaspermroz with the new files it got a bit harder to navigate the `form` folder, would you mind collocating all the files in a single nested folder like `state/*`?

Also the hook name does not meet the convention, can you please rename it to `use-flow-state.ts`?

packages/react/src/components/form/flow.types.ts
```diff
@@ -0,0 +1,90 @@
...
+} from "../../domain/types";
+
+export interface AuthenticatingState {
+  status: "authenticating";
+  context: {
+    config: LoginConfiguration;
+    options?: LoginOptions;
+    attempt: number;
+  };
+  // uiStateMachine? => uiStateMachineFactory(factor) => OTP | TOTP | Password
```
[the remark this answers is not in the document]
ikovic: Looks like the getter is missing? Is this type being used?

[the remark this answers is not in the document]
kaspermroz: It was added in the second PR

packages/react/src/components/form/authenticating/otp.tsx
kaspermroz: Removed in the second PR

packages/react/src/components/form/ui-state-machine.ts
kaspermroz: This was copy-pasted from other place I'm sure, I'll check the behaviour when I move on to QA in another PR

packages/react/src/components/form/flow.ts
kaspermroz: https://github.com/slashid/javascript/pull/239/commits/e6a560b405604d99c62f7ffbb12ad62e2cd06b50

packages/react/src/components/form/ui-state-machine.ts
kaspermroz: SlashIDError is a thing in the ng-evangelion, we don't have such a thing in this repo

ikovic: It is exposed through the `Errors` namespace:

```
import { Errors } from '@slashid/slashid'

new Errors.SlashIDError(// ...
```

kaspermroz:
We decided to take a different approach, this PR is now obsolete. 
New PR: https://github.com/slashid/javascript/pull/249

PR #239, opened 2024-04-18
[SID-1618] TOTP support

kaspermroz:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1618)

Support TOTP as authentication factor.

### QA

Please note that TOTP cannot be uses as the first authentication factor.  The local `dev.tsx` file was modified, so that once you login with Basic form, there will be an option to MFA with TOTP - this is how all the flows should be tested.

**Scenarios**
 - register new Authenticator app
 - login with Authenticator app
 - login with recovery codes
 - test for invalid otp codes
 - try to break it :)

## Checklist


packages/react/src/components/form/authenticating/totp.tsx
```diff
@@ -0,0 +1,283 @@
...
+      {isInputState(uiState) && (
+        <form
+          onSubmit={registerSubmit((e) => {
+            e.preventDefault();
+            uiState.submit(values["totp"]);
+          })}
+          className={styles.otpForm}
+        >
+          <div className={styles.formInner}>
+            {useRecoveryCode ? (
```
kaspermroz, line 163: Note: The design has a gap, there is no screen for logging in with a recovery code, something I overlooked. I figured that it's not super hard to do this now, without reiterating on the design. Let me know WDYT, I think it's sufficient and most importantly - it works. Maybe we can polish it a bit more if needed ✨

packages/react-primitives/src/components/read-only/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+  value: string;
+  id: string;
+  label?: string;
+  as?: "input" | "textarea";
+  rows?: number;
+  copy?: boolean;
+  className?: string;
+};
+
+export function ReadOnly({
```
ikovic, line 16: Can you please specify what is the use case of this component? Some JS-doc style explanations here would be helpful, it's not easy to understand from the code

packages/react/src/components/form/authenticating/authenticating.components.tsx
```diff
@@ -22,6 +23,7 @@ export const BackButton = ({ onCancel }: { onCancel: () => void }) => {
   );
 };
 
+// TODO swap with a generic one
```
ikovic, line 26: Is this going to happen in the same PR?

packages/react/src/components/form/ui-state-machine.ts
```diff
@@ -265,6 +297,111 @@ class PasswordUIStateMachine extends UIStateMachine<PasswordStatus> {
...
+      this.recoveryCodes = recoveryCodes;
+      this.isRegisterFlow = true;
+
+      this.send({
+        type: "sid_ui.totpKeyGenerated",
+        qrCode,
+        uri,
+      });
+    };
+    const totpCodeRequestedHandker = () => {
```
ikovic, line 319: Typo in the name here.

packages/react/src/components/form/ui-state-machine.ts
```diff
@@ -265,6 +297,111 @@ class PasswordUIStateMachine extends UIStateMachine<PasswordStatus> {
...
+
+        this.logInFn(this.context.config, this.context.options)
+          .then((user) => {
+            if (user) {
+              if (this.isRegisterFlow) {
+                this.send({ type: "sid_ui.userAuthenticated", user });
+              } else {
+                this.send({ type: "sid_login.success", user });
+              }
+            } else {
```
ikovic, line 337: An early return in case the user is not returned here would help you get rid of some indentation

kaspermroz:
We decided to take a different approach, this PR is now obsolete. 
New PR: https://github.com/slashid/javascript/pull/249

PR #247, opened 2024-04-24
Write e2e tests for login flows

kaspermroz:
## Description

Fragile test scenarios from the TOTP PR review: https://github.com/slashid/javascript/pull/236#issuecomment-2069128284

## Checklist


packages/tests/e2e/auth.spec.ts
```diff
@@ -60,4 +60,125 @@ test.describe("Authentication", () => {
...
+    await formPage.selectAuthMethod("password");
+    await formPage.enterEmail(testInbox.email);
+    await formPage.submitInitialForm();
+
+    await expect(formPage.authenticatingState).toBeVisible();
+
+    // verify email on register
+    const email = await testInbox.getLatestEmail();
+    expect(email).toBeDefined();
+    if (!email) return;
```
paulo-raca, line 79: What is the different between `expect(email).toBeDefined()` and `if (!email)`?

paulo-raca (approved):

packages/tests/src/slashid.ts
```diff
@@ -0,0 +1,38 @@
+import {
+  PersonsService,
+  OpenAPI,
+  PersonHandleType,
+  PersonCredentialsService,
+} from "../slashid";
+import { hash } from "bcryptjs";
+
+const oid = process.env.E2E_SID_ORG_ID!;
```
ikovic, line 9: This is unsafe as the whole test suite depends on these variables to be set. That is solved by creating a config object that has runtime validation built in, so it throws an error when these variables are not set. We have such an object in `playwright.config.ts`, can you please extract it and use the config here instead of accessing the variables directly?

packages/tests/src/config.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  if (!config.apiURL) {
+    throw new Error("E2E_SID_API_URL us required in .env file");
+  }
+
+  return config;
+}
+
+export const config = createE2ETestConfig();
```
ikovic, line 29: This is missing the other required variables - can you check what happens in `playwright.config`, add the missing variables and use the same config everywhere?

packages/tests/config.ts
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  if (!config.apiURL) {
+    throw new Error("E2E_SID_API_URL us required in .env file");
+  }
+
+  return config;
+}
+
+export const config = createE2ETestConfig();
```
ikovic, line 41: @kaspermroz sorry, I missed this initially when setting this project up and now in the code review again - there are more env vars in the `email` module that require the same treatment:
```
 const config = {
    serverId: process.env.MAILOSAUR_SERVER_ID || "",
    apiKey: process.env.MAILOSAUR_API_KEY || "",
  };

  if (!config.serverId || !config.apiKey) {
    throw new Error("Mailosaur configuration is missing");
  }
```

Can you add those here as well?

kaspermroz: Sure! Sorry I forgot about these too 😬 on it

packages/tests/package.json
```diff
@@ -20,8 +24,11 @@
   "dependencies": {
     "@slashid/slashid": "^3.22.0",
     "@types/uuid": "^9.0.8",
+    "axios": "^1.6.8",
+    "bcryptjs": "^2.4.3",
     "cheerio": "1.0.0-rc.12",
     "dotenv": "^16.4.5",
+    "form-data": "^4.0.0",
```
ikovic, line 31: @kaspermroz this looks accidental, can you check please?

packages/tests/src/email.ts
```diff
@@ -64,11 +56,19 @@ export function createTestInbox({
     return message.subject.match(/\d{6}/)?.[0];
   }
 
+  async function getEmailBySubject(subject: string) {
+    return mailosaurClient.messages.get(config.mailServerId, {
```
ikovic, line 60: Their API is super convenient!

packages/tests/package.json
kaspermroz: This is for the autogenerated client, similarly to `axios`

packages/tests/e2e/auth.spec.ts
```diff
@@ -60,4 +61,92 @@ test.describe("Authentication", () => {
 
     await expect(formPage.successState).toBeVisible();
   });
+
+  test("Reset password flow", async ({ page, context }) => {
```
ikovic, line 65: The top level test suite is `Authentication` - this is actually `Recovery`. Can you add a new top level `describe` called `Recovery` and place this scenario there?

packages/tests/e2e/auth.spec.ts
```diff
@@ -60,4 +61,92 @@ test.describe("Authentication", () => {
...
+
+    const email = await testInbox.getLatestEmail();
+    expect(email).toBeDefined();
+    if (!email) return;
+
+    const otp = "000000";
+
+    await formPage.submitOTP(otp);
+
+    await expect(formPage.errorMessage).toBeVisible();
```
ikovic, line 132: Do you think it would still make sense to submit the correct code to ensure that you actually can log in after submitting an incorrect code at least once?

packages/tests/e2e/auth.spec.ts
```diff
@@ -60,4 +61,92 @@ test.describe("Authentication", () => {
...
+    await formPage.selectAuthMethod("otp_via_email");
+    await formPage.enterEmail(testInbox.email);
+    await formPage.submitInitialForm();
+
+    const email = await testInbox.getLatestEmail();
+    expect(email).toBeDefined();
+    if (!email) return;
+
+    await formPage.promptCTA(); // resend OTP
+    await expect(formPage.emailIcon).toBeVisible();
```
ikovic, line 150: Similar question here, should we await for the second email and make sure a new code was delivered and you could log in? This was broken at least once so IMO it would make sense to do it.

packages/tests/src/pom/form.ts
```diff
@@ -61,4 +78,14 @@ export class FormPage {
       index++;
     }
   }
+
+  async promptCTA() {
```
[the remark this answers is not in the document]
kaspermroz: Yes, I think that does make a lot of sense - I'll make sure that public API of this POM is more explicit

packages/tests/e2e/auth.spec.ts
```diff
@@ -60,4 +61,92 @@ test.describe("Authentication", () => {
...
+
+    await formPage.selectAuthMethod("otp_via_email");
+    await formPage.enterEmail(testInbox.email);
+    await formPage.submitInitialForm();
+
+    const email = await testInbox.getLatestEmail();
+    expect(email).toBeDefined();
+    if (!email) return;
+
+    await formPage.resendOTPCode();
```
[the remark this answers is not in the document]
kaspermroz: I can actually ensure that from within the public API of the POM

packages/tests/e2e/auth.spec.ts
```diff
@@ -60,4 +61,142 @@ test.describe("Authentication", () => {
...
+
+    const beforeResend = new Date(Date.now());
+    await formPage.resendOTPCode();
+    await expect(formPage.emailIcon).toBeVisible();
+
+    const maxRetries = 4;
+    let retries = 0;
+    let newOtp = "";
+
+    while (retries < maxRetries) {
```
ikovic, line 121: If I understand this right, this code block polls the email service until the email with the new OTP code arrives. This seems to be on a lower level of abstraction when compared to the rest of the steps in this test case. How do you feel about extracting this to a function named like `fetchNewOTP` or something?

kaspermroz: Would you have it as a method on `testInbox` then?

ikovic: I think it's very specific to the OTP flow, so I would define that function as close as possible to the actual test case.

ikovic (approved):
Great work @kaspermroz , thanks!

PR #249, opened 2024-04-25
[SID-1618] Add TOTP support

kaspermroz:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1618)

Support TOTP as authentication factor.

## QA
Please note that TOTP cannot be uses as the first authentication factor. The local dev.tsx file was modified, so that once you login with Basic form, there will be an option to MFA with TOTP - this is how all the flows should be tested.

### Scenarios

- register new Authenticator app
- login with Authenticator app
- login with recovery codes
- test for invalid otp codes
- try to break it :)

## Checklist


packages/react/src/components/form/form.tsx
```diff
@@ -110,6 +111,11 @@ export const Form = ({
           {flowState.status === "authenticating" && (
             <FormProvider>{slots.authenticating}</FormProvider>
           )}
+          {flowState.status === "storeRecoveryCodes" && (
+            <FormProvider>
```
ikovic, line 115: Is the `FormProvider` needed here? Looking at `StoreRecoveryCodes` it seems not, but I might be wrong 😄

kaspermroz: Nope it's not, I'll remove it :)

packages/react/src/components/form/authenticating/index.tsx
```diff
@@ -40,28 +50,44 @@ function Wrapper({ children }: { children: React.ReactNode }) {
   );
 }
 
-export const Authenticating = ({ flowState }: Props) => {
+export const Authenticating = ({ flowState }: Pick<Props, "flowState">) => {
   const factor = flowState.context.config.factor;
+  const isLoggingIn = useRef(false);
+
+  const performLogin = () => {
```
ikovic, line 57: `performLogin` is being used as a dependency in `useEffect` so it is safer to create it using `useCallback` to prevent any unwanted effects firing.

ikovic:
## QA notes

### Error handling

Entering an incorrect code after trying to register leads you back to the QR code scanning state. Instead it would be expected to be able to enter the code again.

packages/react/src/components/form/store-recovery-codes/store-recovery-codes.css.ts
```diff
@@ -0,0 +1,12 @@
+import { globalStyle, style } from "@vanilla-extract/css";
+
+export const recoveryCodes = style({
+  marginTop: 24,
+});
+
+globalStyle(`${recoveryCodes} > div`, {
```
ikovic, line 7: This can break, we can work around it this way:
- expose a prop on the `ReadOnlyField` like `fieldClassName`
- then apply it to the child element, regardless of the element type

What do you think?

packages/react/src/components/form/store-recovery-codes/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+          className={styles.recoveryCodes}
+        />
+        <DownloadCodes codes={flowState.context.recoveryCodes} />
+        <Button
+          className={sprinkles({
+            marginTop: "2",
+          })}
+          type="submit"
+          variant="primary"
+          testId="sid-form-authenticating-submit-button"
```
ikovic, line 55: Is this a good test ID? Maybe it should be `sid-form-recovery-codes-submit-button`?

packages/react-primitives/src/components/read-only-field/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const inputRef = useRef<HTMLElement>(null);
+
+  const copyToClipBoard = () => {
+    if (typeof window !== "undefined") {
+      navigator.clipboard.writeText(value);
+    }
+  };
+
+  // prevent horizontal scrolling
+  useEffect(() => {
```
ikovic, line 38: This looks a bit surprising, what is the problem and can we solve it some other way?

kaspermroz: The problem here is that the URI is longer than the input. Even if we do text wrap with CSS, I can still scroll horizontally with Magic mouse or touchpad and make the value disappear. I tried 5 different ways of fixing this with CSS and only that JS snipped helped 🤷‍♂️

ikovic: How about `text-overflow: ellipsis;`?

ikovic: Not sure about UX there, in case someone wants to select and copy the code, they wouldn't have the whole string 🤔 I guess any CSS solution would be preferred but if not possible I guess this is OK

ikovic: Right now, two issues are present:
- there is no feedback on Copy icon click (at least some kind of visual feedback should be present indicating that the value has been copied)
- you cannot manually copy the whole code

ikovic: A nice pattern I saw was temporarily switch the icon from the copy icon to checked icon. Also flash the border or something like that

ikovic (approved):

PR #252, opened 2024-04-29
[Anon fix] Extend the API with anon users instead of mutating it

ikovic:
## Anon users API - extension

In this version of the anonymous users API, anon users are exposed through the additional `anonymous` user property of the `useSlashID` hook.

If we choose this solution, we also need to update the docs as the API is different.


kaspermroz (approved):
LGTM, great job! I think it's the best option we had to tackle anon users in React SDK nicely 😎

PR #256, opened 2024-05-03
[SID-1699] Granular error states: user with a no password set + no reachable handle

ikovic:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1699)

When a user that was created with username only (eg no reachable handles) tries to sign in with a password, they shouldn't be able to set a password in case they haven't had one to start with.

This throws a specific error type that the React SDK picks up and renders the error state of the form accordingly.

## Checklist

### DIFF
--- a/packages/react/src/components/form/form-error.test.tsx
+++ b/packages/react/src/components/form/form-error.test.tsx
@@ -4,6 +4,8 @@ import { TestSlashIDProvider } from "../../context/test-providers";
 import { Form } from "./form";
 import { createTestUser, inputEmail } from "../test-utils";
 import { ConfigurationProvider } from "../../main";
+import { Errors } from "@slashid/slashid";
+import { ERROR_NAMES } from "../../domain/errors";
 
 describe("#Form -> Error state", () => {
   test("should show the error state if login fails", async () => {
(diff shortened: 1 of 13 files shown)


packages/react/src/components/form/form-error.test.tsx
```diff
@@ -0,0 +1,214 @@
...
+    user.click(screen.getByTestId("sid-form-initial-submit-button"));
+
+    await expect(
+      screen.findByTestId("sid-form-error-state")
+    ).resolves.toBeInTheDocument();
+    expect(screen.getByText(testTitle)).toBeInTheDocument();
+  });
+
+  test("should render the recoverNonReachableHandleType error state", async () => {
+    const logInMock = vi.fn(() =>
```
ikovic, line 184: This is incorrect as the `logInMock` should not be called at all

packages/react/src/components/form/form-error.test.tsx
```diff
@@ -0,0 +1,214 @@
...
+      <TestSlashIDProvider sdkState="ready" logIn={logInMock}>
+        <ConfigurationProvider
+          text={{ "error.title.recoverNonReachableHandleType": testTitle }}
+        >
+          <Form />
+        </ConfigurationProvider>
+      </TestSlashIDProvider>
+    );
+
+    inputEmail("valid@email.com");
```
ikovic, line 205: Incorrect, must be `username`

packages/react/src/components/form/form-error.test.tsx
ikovic: Fixed

PR #261, opened 2024-05-21
[SID-1561] Display handle value in authenticating state

kaspermroz:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1561)

Sometimes people make typos when trying to sign up with SlashID. Once you submit the form, you can't see the exact handle you just typed in anymore. We want to add this information, so that people can see if they made a typo in case they don't receive an email or sms.

## Checklist

### DIFF
--- a/packages/react/src/components/form/authenticating/messages.ts
+++ b/packages/react/src/components/form/authenticating/messages.ts
@@ -7,16 +7,20 @@ type AuthenticatingMessageOptions = {
   hasRetried: boolean;
 };
 
+type HandleTextTokens = Partial<
+  Record<"EMAIL_ADDRESS" | "PHONE_NUMBER", string>
+>;
+
 /**
  *  Helper function to extract phone number or email address if either is present in the handle.
  *
  * @param {Handle | undefined} handle Handle used for authentication. Can be undefined (e.g. when
  * authenticating with OIDC).
- * @returns {Record<string,string> | undefined} tokens map or undefined
+ * @returns {HandleTextTokens | undefined} tokens map or undefined
  */
-export function getTokensFromHandle(
+export function getTextTokensFromHandle(
   handle: Handle | undefined
-): Record<string, string> | undefined {
+): HandleTextTokens | undefined {
   if (handle && handle.type === "email_address") {
     return {
       EMAIL_ADDRESS: handle.value,
@@ -35,6 +39,7 @@ export function getTokensFromHandle(
 // TODO add case for password
 export function getAuthenticatingMessage(
   factor: Factor,
+  handle: Handle | undefined,
   { isSubmitting, hasRetried }: AuthenticatingMessageOptions = {
     isSubmitting: false,
     hasRetried: false,
@@ -42,6 +47,7 @@ export function getAuthenticatingMessage(
 ): {
   title: TextConfigKey;
   message: TextConfigKey;
+  tokens?: HandleTextTokens;
 } {
   switch (factor.method) {
     case "oidc":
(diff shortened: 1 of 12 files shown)


packages/react/src/components/form/form-customisation.test.tsx
```diff
@@ -92,19 +95,21 @@ describe("#Form - customisation", () => {
...
-                <div data-testid="custom-error-function">
-                  <h1>{context.error.message}</h1>
-                  <button onClick={() => retry("retry")}>Retry</button>
-                  <button onClick={cancel}>Cancel</button>
-                </div>
-              )}
-            </Form.Error>
-          </Slot>
-        </Form>
+        <TestTextProvider text={TEXT}>
```
kaspermroz, line 98: The interpolation tokens functionality depends on the `text` object in the context, otherwise it breaks

packages/react/src/components/form/authenticating/messages.ts
```diff
@@ -1,19 +1,48 @@
...
+/**
+ *  Helper function to extract phone number or email address if either is present in the handle.
+ *
+ * @param {Handle | undefined} handle Handle used for authentication. Can be undefined (e.g. when
+ * authenticating with OIDC).
+ * @returns {Record<string,string> | undefined} tokens map or undefined
+ */
+export function getTokensFromHandle(
+  handle: Handle | undefined
+): Record<string, string> | undefined {
```
ikovic, line 19: We can be more precise when describing the return type as we know the shape of the object

packages/react/src/components/form/authenticating/messages.ts
```diff
@@ -1,19 +1,48 @@
...
 };
 
+/**
+ *  Helper function to extract phone number or email address if either is present in the handle.
+ *
+ * @param {Handle | undefined} handle Handle used for authentication. Can be undefined (e.g. when
+ * authenticating with OIDC).
+ * @returns {Record<string,string> | undefined} tokens map or undefined
+ */
+export function getTokensFromHandle(
```
ikovic, line 17: `token` is an overloaded term, maybe this would be more accurate: `getTextTokensFromHandle`?

packages/react/src/components/text/constants.ts
```diff
@@ -86,28 +86,30 @@ export const TEXT = {
   "authenticating.confirm": "Confirm",
   "authenticating.downloadCodes": "Download codes",
   "authenticating.message.webauthn":
-    "If you are registering for the first time, you will receive an email to verify your email address.",
+    "If you are registering for the first time, you will receive an email to verify your email address {{EMAIL_ADDRESS}}.",
```
ikovic, line 89:
```suggestion
    "If you are registering for the first time, you will receive an email at {{EMAIL_ADDRESS}} to verify your email address.",
```

packages/react/src/components/form/authenticating/otp.tsx
```diff
@@ -50,12 +50,13 @@ export const OTPState = ({ flowState, performLogin }: Props) => {
...
   const submitInputRef = useRef<HTMLInputElement>(null);
 
-  const factor = flowState.context.config.factor;
+  const { factor, handle } = flowState.context.config;
   const hasRetried = flowState.context.attempt > 1;
   const { title, message } = getAuthenticatingMessage(factor, {
     isSubmitting: formState === "submitting",
     hasRetried,
   });
+  const tokens = getTokensFromHandle(handle);
```
ikovic, line 59: It seems strange that the message tokens are returned separately from the message. Would it make sense if `getAuthenticating` message returns both the message and the related tokens?

ikovic (approved):

PR #264, opened 2024-06-04
[SID-1742] React SDK: Send FlowCancelled event when the flow is cancelled

kaspermroz:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1742)

This PR resolves the issue of flow cancellation when the polling doesn't stop if you cancel the flow,

## Checklist

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -65,7 +65,7 @@
   },
   "devDependencies": {
     "@faker-js/faker": "^8.0.2",
-    "@slashid/slashid": "3.25.0",
+    "@slashid/slashid": "3.26.1-cancel-flow.2",
     "@storybook/addon-essentials": "7.6.19",
     "@storybook/addon-interactions": "7.4.0",
     "@storybook/addon-links": "7.4.0",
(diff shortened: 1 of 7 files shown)


packages/react/package.json
```diff
@@ -65,7 +65,7 @@
   },
   "devDependencies": {
     "@faker-js/faker": "^8.0.2",
-    "@slashid/slashid": "3.25.0",
+    "@slashid/slashid": "3.26.1-cancel-flow.2",
```
kaspermroz, line 68: Replace this with a latest release once it's there

ikovic (approved):

packages/react/package.json
```diff
@@ -65,7 +65,7 @@
   },
   "devDependencies": {
     "@faker-js/faker": "^8.0.2",
-    "@slashid/slashid": "3.25.0",
+    "@slashid/slashid": "3.26.1",
```
ikovic, line 68: Please remember to update the peer dependency too, as this won't work unless the core SDK is installed as well

kaspermroz: Good point! Done https://github.com/slashid/javascript/pull/264/commits/20f11d3154f0fffc7ebe42ad4611e6f22bc9c763

PR #273, opened 2024-06-10
React: expose authentication form state

jake-slashid:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1764)

This PR exposes the internal `<Authenticating>` state for advanced use cases where rendering individual states manually is necessary.

Public interface looks like this:
```
import { States } from '@slashid/react'

<States.Authenticating flowState={...} />
```

### DIFF
--- a/packages/react/src/components/form/states.tsx
+++ b/packages/react/src/components/form/states.tsx
@@ -1,0 +1,14 @@
+import clsx from "clsx";
+import * as styles from "./form.css";
+import { AuthenticatingProps, Authenticating as AuthenticatingState } from "./authenticating";
+import { ConfigurationProvider } from "../../context/config-context";
+
+export const Authenticating = (props: AuthenticatingProps) => {
+  return (
+    <div className={clsx("sid-form", styles.form)}>
+      <ConfigurationProvider>
+        <AuthenticatingState flowState={props.flowState} />
+      </ConfigurationProvider>
+    </div>
+  )
+}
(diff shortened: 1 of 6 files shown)


packages/react/src/components/form/states.tsx
```diff
@@ -0,0 +1,14 @@
+import clsx from "clsx";
+import * as styles from "./form.css";
+import { AuthenticatingProps, Authenticating as AuthenticatingState } from "./authenticating";
+import { ConfigurationProvider } from "../../context/config-context";
+
+export const Authenticating = (props: AuthenticatingProps) => {
+  return (
+    <div className={clsx("sid-form", styles.form)}>
+      <ConfigurationProvider>
```
ikovic, line 9: Have you tried using this in the Console ready? Looking only at the code here I'm pretty sure that this will make passing customisation props impossible.

The Console renders a ConfigurationProvider which it uses to inject the customisation props in the UI. So rendering one here breaks that.

Ideally we'd just expose the regular Authenticating state component and wrap it with required context there.

jake-slashid: Okay sure, I was trying to make the interface as simple as possible which is why it was baked in here

packages/react/src/components/form/states.tsx
```diff
@@ -0,0 +1,11 @@
+import clsx from "clsx";
+import * as styles from "./form.css";
+import { AuthenticatingProps, Authenticating as AuthenticatingState } from "./authenticating";
+
+export const Authenticating = (props: AuthenticatingProps) => {
```
ikovic, line 5: It's quite likely that over time we'll forget why this is here, maybe just a comment explaining the use case of this module?

ikovic:
@jake-slashid needs a changeset, can you add one please?

ikovic (approved):

kaspermroz (approved):

PR #279, opened 2024-06-13
[SID-1773] React: use css to support line breaks

jake-slashid:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1773)

Same as https://github.com/slashid/javascript/pull/277 but using CSS only


ikovic (approved):
Don't forget to add a changeset :D

kaspermroz (approved):

PR #283, opened 2024-06-20
[SID-1762] React: default to 'localStorage' storage when anonymous users enabled

jake-slashid:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1762)

If anonymous users is turned on but storage is left default, it will be `memory` which means the anonymous users is lost each time the page refreshes.

- Set storage to `localStorage` by default if anonymous users enabled
- When new default is invoked, emit warning to set storage explicitly
- Warn if storage is explicitly set to `memory` while anonymous users enabled

### DIFF
--- a/packages/react/src/context/slash-id-context.tsx
+++ b/packages/react/src/context/slash-id-context.tsx
@@ -105,15 +105,23 @@ SlashIDContext.displayName = "SlashIDContext";
 
 const STORAGE_TOKEN_KEY = "@slashid/USER_TOKEN";
 
-const createStorage = (storageType: StorageOption) => {
+const createStorage = (storageType: StorageOption, options: { anonymousUsersEnabled: boolean }) => {
   switch (storageType) {
     case "memory":
+      if (options.anonymousUsersEnabled) {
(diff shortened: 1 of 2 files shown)


ikovic:
@jake-slashid I like the runtime change, but I feel it would still be well worth it to also prevent this during static checks - having a union type discriminated by the `anonymousUsersEnabled` property on the `SlashIDProvider` could filter out `memory` as an option when that flag is raised.

packages/react/src/context/slash-id-context.tsx
```diff
@@ -105,15 +105,23 @@ SlashIDContext.displayName = "SlashIDContext";
 
 const STORAGE_TOKEN_KEY = "@slashid/USER_TOKEN";
 
-const createStorage = (storageType: StorageOption) => {
+const createStorage = (storageType: StorageOption, options: { anonymousUsersEnabled: boolean }) => {
```
ikovic, line 108: IMO this must be handled one level above, as the SlashID context provider is the owner of this business logic.

PR #286, opened 2024-07-17
React: use type=email for email handle input

jake-slashid:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1847)

When entering handle, the inputs have the following types:
- Phone number: `tel`
- Username: `text`
- Email: `text`

This PR changes email to type `email`, and explicitly sets phone number to type `tel` (though, this is the default already).

The intent is better semantics, and to allow better integration with browsers, and other tools depending on the field being correctly typed.

### DIFF
--- a/.changeset/wicked-spies-deliver.md
+++ b/.changeset/wicked-spies-deliver.md
@@ -1,0 +1,5 @@
+---
+"@slashid/react": minor
+---
+
+Set input="email" for email handle input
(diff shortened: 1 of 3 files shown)


jake-slashid:
This introduces a new validation behaviour from the browser, we need to spend a little more time handling that better.

kaspermroz (approved):

ikovic (approved):

PR #291, opened 2024-08-30
[SID-1935] Pre-select last used auth method

kaspermroz:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1935)

We already have a functionality to store last used handle. This PR adds another configuration property to store last used authentication factor and pre-select it in the dropdown (works for non-oidc factors only).

## Checklist

### DIFF
--- a/.changeset/hot-rules-mix.md
+++ b/.changeset/hot-rules-mix.md
@@ -2,4 +2,4 @@
 "@slashid/react": minor
 ---
 
-Add possibility to store last used authentication factor and pre-select iton next login
+Add possibility to store last used authentication factor and pre-select it on next login
--- a/.changeset/hot-rules-mix.md
+++ b/.changeset/hot-rules-mix.md
@@ -1,0 +1,5 @@
+---
+"@slashid/react": minor
+---
+
+Add possibility to store last used authentication factor and pre-select iton next login
--- a/packages/react/src/hooks/use-last-factor.ts
+++ b/packages/react/src/hooks/use-last-factor.ts
@@ -5,7 +5,8 @@ import { useConfiguration } from "./use-configuration";
 import { useSlashID } from "./use-slash-id";
 import { Factor } from "@slashid/slashid";
 
-export const STORAGE_LAST_FACTOR_KEY = "@slashid/LAST_FACTOR";
+export const STORAGE_LAST_FACTOR_KEY = (oid: string) =>
+  `@slashid/LAST_FACTOR/${oid}`;
 
 type UseLastFactor = () => {
   lastFactor: Factor | undefined;
@@ -25,7 +26,9 @@ export const useLastFactor: UseLastFactor = () => {
     }
 
     try {
-      const storedFactor = window.localStorage.getItem(STORAGE_LAST_FACTOR_KEY);
+      const storedFactor = window.localStorage.getItem(
+        STORAGE_LAST_FACTOR_KEY(sid?.oid ?? "")
+      );
       if (!storeLastFactor || !storedFactor) {
         return undefined;
       }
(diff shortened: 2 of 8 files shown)


packages/react/src/hooks/use-last-factor.ts
```diff
@@ -0,0 +1,73 @@
+import { useEffect, useMemo, useCallback } from "react";
+import { isBrowser } from "@slashid/react-primitives";
+import { Handle, isFactor } from "../domain/types";
+import { useConfiguration } from "./use-configuration";
+import { useSlashID } from "./use-slash-id";
+import { Factor } from "@slashid/slashid";
+
+export const STORAGE_LAST_FACTOR_KEY = "@slashid/LAST_FACTOR";
```
ikovic, line 8: Should this be prefixed with the organization ID? Otherwise if you use different orgs (especially in the hosted login page) you are likely to get an incorrect factor

packages/react/src/hooks/use-last-factor.ts
```diff
@@ -0,0 +1,73 @@
...
+  const { storeLastFactor } = useConfiguration();
+  const { sid } = useSlashID();
+
+  const lastFactor = useMemo(() => {
+    if (!isBrowser()) {
+      return undefined;
+    }
+
+    try {
+      const storedFactor = window.localStorage.getItem(STORAGE_LAST_FACTOR_KEY);
```
jake-slashid, line 28: Don't we have an abstraction for storage? 🤔

kaspermroz: not really, we have a helper to create a storage for `tokenStorage`, other than that we're using local storage if possible

packages/react/src/hooks/use-last-factor.ts
kaspermroz: good point!

packages/react/src/hooks/use-last-factor.ts
```diff
@@ -0,0 +1,79 @@
...
+    }
+  }, [storeLastFactor, sid]);
+
+  const handler = useCallback(
+    ({ authenticationFactor }: SuccessEvent) => {
+      if (!isBrowser()) {
+        return;
+      }
+
+      if (!isFactor(authenticationFactor)) {
```
ikovic, line 48: Why do we need to make this check? I would assume that the `SuccessEvent` would always have this value 🤔

packages/react/src/hooks/use-last-factor.ts
```diff
@@ -0,0 +1,79 @@
...
+
+type UseLastFactor = () => {
+  lastFactor: Factor | undefined;
+};
+
+type SuccessEvent = {
+  authenticationFactor: Handle | undefined;
+};
+
+export const useLastFactor: UseLastFactor = () => {
```
ikovic, line 19: Can we annotate the return value of this hook?

kaspermroz: It is annotated by `UseLastHandle` type

packages/react/src/hooks/use-last-factor.ts
kaspermroz: I mean it's just a bit of defensive programming, very unlikely that there's some garbage in there but does it hurt to have it here?

ikovic (approved):

.changeset/hot-rules-mix.md
```diff
@@ -0,0 +1,5 @@
+---
+"@slashid/react": minor
+---
+
+Add possibility to store last used authentication factor and pre-select iton next login
```
kaspermroz, line 5:
```suggestion
Add possibility to store last used authentication factor and pre-select it on next login
```

PR #292, opened 2024-08-30
[SID-1933] React SDK: Display Contact support CTA as secondary button

kaspermroz:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1933)

Make the Contact support CTA more visible.

Screenshot 2024-08-30 at 16 42 02

## Checklist

### DIFF
--- a/packages/react/src/components/form/error/index.tsx
+++ b/packages/react/src/components/form/error/index.tsx
@@ -175,7 +175,7 @@ function ContactSupportPrompt() {
 
   return (
     <>
-      <Divider>{text["initial.divider"]}</Divider>
+      <Divider>{text["error.divider"]}</Divider>
       <div
         className={clsx("sid-form-error-contact-support", styles.supportPrompt)}
         data-testid="sid-form-error-support-prompt"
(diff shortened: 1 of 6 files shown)


jake-slashid (approved):

packages/react/src/components/form/error/index.tsx
```diff
@@ -173,20 +174,24 @@ function ContactSupportPrompt() {
...
-      data-testid="sid-form-error-support-prompt"
-    >
-      <span>{text["error.contactSupport.prompt"]}</span>
-      <a
-        className={styles.supportCta}
-        target="_blank"
-        href={supportURL}
-        rel="noreferrer"
+    <>
+      <Divider>{text["initial.divider"]}</Divider>
```
ikovic, line 178: We shouldn't reuse the text tokens based on value. This has a different context than the `initial` state - please add another text token.

kaspermroz: oh sorry I missed that it's `initial`, will do!

PR #298, opened 2024-09-19
[SID-1966] Render the Resend prompt immediately, but block it with a counter

ikovic:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-1966)

Some end users are spamming the `resend` button, so they end up in a situation where they receive multiple emails and they don't know which code to use then. We now render the `resend` button immediately, but it is blocked behind a counter which only allows clicks every 30 seconds.

## Checklist


packages/react/src/components/form/authenticating/index.tsx
```diff
@@ -20,9 +21,11 @@ import * as styles from "./authenticating.css";
 import { Children, useCallback, useEffect, useRef, useState } from "react";
 import { TOTPState } from "./totp";
 import { Delayed } from "@slashid/react-primitives";
-import { BASE_RETRY_DELAY_MS } from "./authenticating.constants";
 import { useInternalFormContext } from "../internal-context";
 import { AuthenticatingState } from "../flow";
+import { TIME_MS } from "../types";
+
+const DELAY_BEFORE_RETRY = TIME_MS.second * 30;
```
kaspermroz, line 28: Previously it was 2 seconds, now we're changing it to 30 (which I guess makes sense), just curious if this number is arbitrary or was it an outcome of some request/conversation?

packages/tests/playwright.config.ts
```diff
@@ -4,7 +4,7 @@ import { config as envConfig } from "./config";
 
 function createPlaywrightConfig(): PlaywrightTestConfig {
   return {
-    timeout: 16 * 1000,
+    timeout: 45 * 1000,
```
kaspermroz, line 7: Is this related to the timeout increase? If yes, then I'd mock timers instead of increasing the timeout 🤔

ikovic: Of course. Problem is, it would blow up the scope of this work. These are not unit tests so we don't have that level of control.

E2E tests test apps which depend on the React SDK. Multiple boundaries are in place so we would need a system for injecting a test config when tests are being run. This would likely take a lot of time so I went with the easy way this time.

packages/react/src/components/form/authenticating/index.tsx
kaspermroz: Got it!

packages/tests/playwright.config.ts
kaspermroz: Understood, we should at least make a note of this because it directly makes the CI slower 🥶

kaspermroz (approved):

PR #308, opened 2024-11-15
[SID-2019] Process the authn context update event

ikovic:
## Description

[YouTrack issue](https://todo.irdcorp.dev/issue/SID-2019)

It is possible to miss events published by the core SDK due to race conditions. Having an event buffer makes this less likely to happen as it will replay the events that happened prior the subscription.

With this PR the React SDK will be able to buffer events that were published before any listeners had any chance to process these events. It uses that mechanism to listen to the new event type that establishes the authentication context upon processing the related challenge.

This is a big change, since previously the authentication context was established by the values submitted in the initial step. Now that this can be changed using a sync webhook,  we need to wait for the core SDK to tell us what the auth context it.

## Checklist

### DIFF
--- a/packages/react/src/components/form/error/error.test.tsx
+++ b/packages/react/src/components/form/error/error.test.tsx
@@ -332,7 +332,6 @@ describe("#Form -> Error state -> Special error cases", () => {
     inputUsername("non-reachable");
 
     user.click(screen.getByTestId("sid-form-initial-submit-button"));
-    console.log("clicked");
 
     // go to authenticating state
     await expect(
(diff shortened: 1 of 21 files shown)


packages/react/src/components/form/error/error.test.tsx
```diff
@@ -320,6 +332,7 @@ describe("#Form -> Error state -> Special error cases", () => {
     inputUsername("non-reachable");
 
     user.click(screen.getByTestId("sid-form-initial-submit-button"));
+    console.log("clicked");
```
kaspermroz, line 335: Guess we don't need that

kaspermroz (approved):

packages/react/src/components/form/error/error.test.tsx
ikovic: resolved, thanks

jake-slashid (approved):
Looks very clean, nice work!

PR #312, opened 2024-12-12
React SDK: Fix Radix primitives

kaspermroz:
## Description

## Checklist

### DIFF
--- a/packages/react-primitives/src/components/dropdown/dropdown.css.ts
+++ b/packages/react-primitives/src/components/dropdown/dropdown.css.ts
@@ -39,7 +39,7 @@ export const label = style({
   color: publicVariables.input.label.color,
   position: "absolute",
   top: 12,
-  left: 16,
+  left: 17,
 });
 
 export const input = style({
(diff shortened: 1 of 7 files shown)


ikovic (approved):

packages/react-primitives/src/components/dropdown/index.tsx
```diff
@@ -48,7 +50,10 @@ export const Dropdown: React.FC<Props> = ({
   );
 
   return (
-    <div className="sid-dropdown">
+    <div className={clsx("sid-dropdown beta", styles.wrapper)}>
```
ikovic, line 53: We probably need to remove beta from here, right?

packages/react-primitives/src/components/dropdown/dropdown.css.ts
```diff
@@ -38,6 +37,9 @@ export const label = style({
   fontSize: theme.font.size.xs,
   fontWeight: theme.font.weight.semibold,
   color: publicVariables.input.label.color,
+  position: "absolute",
```
ikovic, line 40: Can we avoid this? Are we sure it's going to look ok in all circumstances?

kaspermroz: Not sure, part of the fix was moving the `<label>` outside of the `<Select.Root>`, but with this UI it's inside of the Select box

PR #315, opened 2024-12-17
React SDK: Add line-height for select label

kaspermroz:
## Checklist


packages/react-primitives/src/components/dropdown/dropdown.css.ts
```diff
@@ -36,6 +36,7 @@ export const trigger = style({
 export const label = style({
   fontSize: theme.font.size.xs,
   fontWeight: theme.font.weight.semibold,
+  lineHeight: "118%",
```
kaspermroz, line 39: This value is consistent with input labels

jake-slashid (approved):

PR #319, opened 2025-01-12
[HRD-Console] React SDK implementation

kaspermroz:
## Description

In this PR:
 - create new SlashID state `__orgSwitchingState`
 - create new component `<OrgSwitchingForm>`

## Checklist


packages/react/package.json
```diff
@@ -60,7 +60,7 @@
   },
   "devDependencies": {
     "@faker-js/faker": "^8.0.2",
-    "@slashid/slashid": "3.28.0",
+    "@slashid/slashid": "3.28.1-hrd-console.0",
```
kaspermroz, line 63: Use non-beta version once it's released

packages/react/src/components/form/org-switching/org-switching-form.tsx
```diff
@@ -0,0 +1,98 @@
...
+    handleValue: undefined,
+    flag: undefined,
+  });
+
+  const { status } = flowState;
+
+  const defaultSlots = useMemo(() => {
+    const slots = {
+      footer: showBanner ? <Footer /> : null,
+      // initial: status === "initial" ? <Initial /> : undefined,
```
ikovic, line 57: leftover

packages/react/src/components/form/org-switching/org-switching-form.tsx
```diff
@@ -0,0 +1,98 @@
...
+export type Props = ConfigurationOverridesProps & {
+  className?: string;
+  onSuccess?: CreateFlowOptions["onSuccess"];
+  onError?: CreateFlowOptions["onError"];
+  middleware?: LoginOptions["middleware"];
+  children?: Slots<"authenticating" | "success" | "error" | "footer">; // TS does not enforce this, but it is used for documentation
+};
+
+/**
+ * Render a form that can be used to sign in or sign up a user.
```
ikovic, line 31: This description is outdated now

packages/react/src/components/form/org-switching/org-switching-form.tsx
```diff
@@ -0,0 +1,98 @@
...
+    };
+
+    return slots;
+  }, [status, showBanner, flowState]);
+
+  const slots = useSlots({ children, defaultSlots });
+
+  return (
+    <InternalFormContext.Provider
+      value={{
```
ikovic, line 72: This will break downstream rendering optimization (in case it is implemented). When setting context values it is better to memoize them first, otherwise a new object is passed every time, causing the context consumers to render needlessly.

packages/react/src/components/form/org-switching/useOrgSwitchingFlowState.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { useSlashID } from "../../../hooks/use-slash-id";
+import {
+  Flow,
+  createFlow,
+  FlowState,
+  CreateFlowOptions,
+} from "./org-switching-flow";
+
+export function useFlowState(opts: CreateFlowOptions = {}) {
```
ikovic, line 11: can you update the name as well?

ikovic:
@kaspermroz can you add a test suite for this flow? Ideally one case for the path when authentication is not required and one with authentication. I think you could potentially do it without mocking any network requests as it is possible to mock the core SDK so we could do it at that level 🤔

packages/react/src/components/form/org-switching/org-switching-flow.ts
```diff
@@ -0,0 +1,455 @@
...
+          ),
+          e
+        );
+        break;
+      case "sid_cancel":
+        if (typeof cancelFn === "function") {
+          cancelFn();
+        }
+
+        // setState(createInitialState(send), e);
```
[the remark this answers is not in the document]
kaspermroz: Good catch, the cancellation of this flow is tricky in couple ways 🤔 I need to think about it, but I'm kind of leaning towards showing error state / throwing an error so that you can give control elsewhere, we might need this in the Console to cancel the

packages/react/src/dev.tsx
```diff
@@ -205,21 +206,18 @@ const BasicForm = () => {
...
-              onError={(error, context) =>
-                console.log("onError", { error, context })
-              }
-            />
-          </LoggedOut>
-        </>
-      </SlashIDLoaded>
+      <>
+        <LoggedIn>
+          <OrgSwitchingForm oid="758e443e-d2c1-2fcc-28b5-0b94c33b5a7c" />
```
ikovic, line 211: Please use env vars, makes it easier for anyone working on this later 😃

packages/react/src/context/slash-id-context.tsx
```diff
@@ -180,11 +226,32 @@ export const SlashIDProvider = ({
...
+      });
+
+      let newToken: string;
+
+      // check if a valid token for new org is in storage
+      const newOidToken = storageRef.current?.getItem(
+        `@slashid/USER_TOKEN/${newOid}`
+      );
+      const isNewOidTokenValid =
+        newOidToken && (await validateToken(newOidToken));
```
ikovic, line 241: Along with validating the token, should we also decode it and check the org ID to ensure it really is a token for the given org? Not sure if that would break some behaviour though 🤔 for example how a token for the root dashboard org can be used for suborgs too. Probably better to skip that check 🤔

packages/react/src/context/slash-id-context.tsx
```diff
@@ -200,7 +267,7 @@ export const SlashIDProvider = ({
         __switchOrganizationInContext({ oid: newUser.oid });
       }
     },
-    [state, __switchOrganizationInContext, oid]
+    [state, __switchOrganizationInContext, oid, STORAGE_TOKEN_KEY]
```
ikovic, line 270: Why is `STORAGE_TOKEN_KEY` a dependency here?

packages/react/src/context/slash-id-context.tsx
```diff
@@ -158,7 +176,16 @@ export const SlashIDProvider = ({
...
+  const [orgSwitchingState, setOrgSwitchingState] = useState<OrgSwitchingState>(
+    {
+      state: "idle",
+    }
+  );
 
+  /**
+   * Storage key for the User token (current organization).
+   */
+  const STORAGE_TOKEN_KEY = useMemo(() => NEW_STORAGE_TOKEN_KEY(oid), [oid]);
```
ikovic, line 188: This is a bit confusing as `STORAGE_TOKEN_KEY` is not a constant value, I'd suggest camelcase here

packages/react/src/context/slash-id-context.tsx
ikovic: I understand why now, got confused with all caps. Probably better to use camelcase

ikovic:
@kaspermroz I think this is on track with what we discussed, can you release a beta and see how this works when used in the console?

packages/react/src/components/form/org-switching/org-switching-flow.ts
```diff
@@ -0,0 +1,458 @@
...
+
+    observers.forEach((o) => o(state, changeEvent));
+  }
+
+  let state: FlowState = createAuthenticatingState(
+    send,
+    {
+      config: {
+        factor: {
+          method: "email_link",
```
ikovic, line 318: This is hardcoded, does it have to be here?

packages/react/src/components/form/org-switching/org-switching-flow.ts
```diff
@@ -0,0 +1,458 @@
...
+        }
+
+        break;
+      default:
+        break;
+    }
+  }
+
+  // provide the flow API - interact with the flow using the state object, subscribe and unsubscribe to state changes
+  return {
```
ikovic, line 436: Most of the object we return from this factory function is the same as the other flow, so it could be a good candidate for refactor.

packages/react/src/components/form/org-switching/org-switching-flow.ts
```diff
@@ -0,0 +1,458 @@
...
+
+  // each history entry contains a state and the event that triggered the transition to that state
+  const history: HistoryEntry[] = [{ state, event: { type: "sid_init" } }];
+
+  const { onSuccess, onError } = opts;
+
+  // notify subscribers every time the state changes
+
+  async function transition(e: Event) {
+    switch (e.type) {
```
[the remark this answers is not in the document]
ikovic: For example, instead of a switch we could pass in an object or a map that maps handlers to event names to make this reusable.

packages/react/src/context/slash-id-context.tsx
```diff
@@ -243,7 +310,17 @@ export const SlashIDProvider = ({
       return;
     }
 
-    storageRef.current?.removeItem(STORAGE_TOKEN_KEY);
+    // search storage for user tokens
+    Object.entries(storageRef.current!).forEach(([key, maybeToken]) => {
+      // check if the storage entry is a user token
+      if (key.startsWith(LEGACY_STORAGE_TOKEN_KEY)) {
```
ikovic, line 316: Wouldn't this remove the new tokens too?

kaspermroz: yes, that's the point here - remove all the tokens we have in storage for log out

ikovic (approved):

